PR #4, opened 2014-09-28
Add initial ./src file structure

matthiasbeyer:
This PR adds the initial structure for `./src`, including files which only contain the license header by now.

### DIFF
--- a/src/action/manager.c
+++ b/src/action/manager.c
@@ -1,0 +1,27 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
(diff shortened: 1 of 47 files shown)


neithernut:
ACK.

matthiasbeyer:
The PR changed a bit, as `src/util/attributes.h` was already added. Please review again.

neithernut:
I just noticed that the include-guards are missing from the header files.

matthiasbeyer:
Yes, I don't want to write _any_ source in this PR. And of course, include guards are source!

matthiasbeyer:
If this should be included in this PR, I'll add it... if you want me to!

neithernut:
k, then. I'll open an issue for that.
ACK.

matthiasbeyer:
Saw my latest comment before commenting?

neithernut:
Ah, nay...
Yes, please add the include guards to the PR. It's much cleaner this way.

neithernut:
Could you also add inclusions of the header files from the corresponding source files? (assume `src` as an include directory).

matthiasbeyer:
Nope, because I don't know about the implementation.

I wouldn't like to add it, even if I knew about how things get implemented. I consider it as better if the person who actually implements it has to deal with "what must I include for this".

neithernut:
Let me rephrase that: given a file `a.h` and a file `a.c`. Can you add an inclusion of `a.h` in `a.c`?
We will need _those_ definitely.

matthiasbeyer:
Of course. I will add this.

neithernut:
I prepared a commit. It's availible in my public repository. You can fetch, review and squash it.
(bash ftw.)

matthiasbeyer:
Nice! You scripted it in bash? 

Could you re-do the commit and add a blank line between the `include` and the last comment line? Would be a nicer format, IMHO.

neithernut:
Done.

matthiasbeyer:
I'm ready for squashing, are you?

neithernut:
Yep.

matthiasbeyer:
Done, I wouldn't squash more. Feel free to merge.

Issue #7, opened 2014-09-28
Create an organisation/project logo


matthiasbeyer:
It would be pretty neat if we could add an organizations/project logo.

neithernut:
My first take on this is uploaded at http://webuser.hs-furtwangen.de/~ganzjuli/waysome.svg. I didn't got the hair quite right, but it's a start.
Comments?

matthiasbeyer:
Nice. I would make the arms a bit longer, and I would put something on the shirt (like a crossed-out X or something :-P )

neithernut:
As already stated somewhere else, we may create some flavored versions (HFU logo and stuff).
This is the basic logo without anything on the shirt.

matthiasbeyer:
Ok. I really like it, ... but maybe we should add something like trouser pockets (just hinted), so it doesn't look like a pyjama!

matthiasbeyer:
And, well, the hair looks a bit iffy...

neithernut:
I already pointed out the latter one. I'll try to fix that issues until 2014-09-29T16:00CEST.

matthiasbeyer:
Also, if this logo was created by something like metapost or so, we could add a repository containing the logo sources.

I would welcome metapost (or racket?) sources to generate the logo! Would be kind of fun!

neithernut:
The logo was done in inkscape. Also: svgs are XML. As such, they _are_ editable. (I once "wrote" an svg for a scale.)

matthiasbeyer:
XML. Yuck. But hey, Inkscape is a great tool, so I don't blame you for using it!

neithernut:
XML is OK for document and other file formats...
It's just misused in a lot of places.

xVin:
If you want, you can leave graphics stuff to me. I'd work in gimp, though.

neithernut:
The base really should be an svg. It's a lot easier to edit this way. I'm not sure whether gimp can handle vector graphics properly. If you want to, you can download the SVG and mock around with it. As stated above, I'm using [inkscape](http://www.inkscape.org).
However, _after_ the basic figure is done, we may, of course, use raster graphics (e.g. gimp) to fancy things up.

---

I already redid the hair of the figure: http://webuser.hs-furtwangen.de/~ganzjuli/waysome-2.svg
I can add pockets. If everyone is satisfied, I'd add the svg to the repo.

xVin:
[...]

PR #21, opened 2014-09-29
Add initial version of waysome logo

neithernut:
This PR adds the waysome logo to the repository.
This solves issue #7 and uses the logo introduced in the discussion of this issue. However, the version added in this PR is stripped of all inkscape-specific stuff and contains additional metadata (title and author).


matthiasbeyer:
NOT ACK.

As stated in #7 , binary data shouldn't be added. Well, this is XML, but as you don't hand-edit it, there is no point in version-tracking it.

We could, of course, add a repository "logo" or something.

neithernut:
There _is_ a point in version tracking this!
If you, for example, add a path or a figure, that path will turn up as a number of new lines.
All the work flows used for source code do make sense for this image too. E.g. you can resolve merge conflicts just as in any other text file.

matthiasbeyer:
Please resubmit on `develop`.

PR #33, opened 2014-09-29
Fix build env

neithernut:
Fix typos which did not trigger failures on this machine because I was too stupid to clear the cache before testing the build in the end.


TheNeikos:
Indeed, they can be nicely overlooked.

ACK as well

PR #38, opened 2014-09-29
Thread safeness for ws_object type

matthiasbeyer:
work in progress. The actual type implementation is not thread safe
by now.

### DIFF
--- a/src/objects/object.c
+++ b/src/objects/object.c
@@ -229,30 +229,6 @@ ws_object_unlock_write(
     return 0 == pthread_rwlock_unlock(&self->rw_lock);
 }
 
-bool
-ws_object_is_locked(
-    struct ws_object const* const self
-) {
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
I consider the `ws_object` type implementation as ready now.

src/objects/object.c
```diff
@@ -133,8 +174,18 @@ void
...
-    /** @todo implement */
-    return;
+    if (self) {
+        wrlock(self);
+        self->ref_counting.refcnt--;
+        self->id->unref_callback(self);
+
+        if (self->ref_counting.refcnt == 0) {
+            self->id->free_callback(self);
+            free(self);
```
neithernut, line 187: The rwlock must be deinitialized before freeing the memory.

neithernut:
ACK.

Issue #46, opened 2014-09-29
Add basic main loop


neithernut:
We need a prelimary main loop implementation.
For now, it should loop until the process receives `SIGTERM` or `SIGINT`.
Modules will add their own watchers and handlers.

matthiasbeyer:
@neithernut close?

PR #42, opened 2014-09-29
Add build env libdrm

neithernut:
Add `libdrm` to build dependencies.
Targets issue #41.

PR #48, opened 2014-09-30
Added simple main loop with signal handlers

TheNeikos:
This is for #46.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -25,10 +25,38 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <stdio.h>
+#include <stdlib.h>
+#include <ev.h>
+
+static void
+handle_sig(
+    struct ev_loop* loop,
+    struct ev_signal* w,
+    int revent
+) {
+    // TODO Do proper logging here
+    fprintf(stderr, "Caught signal %d\n", w->signum);
+    ev_unloop(loop, EVUNLOOP_ALL);
+}
+
 int
 main(
     int argc,
     char** argv
 ) {
+
+    struct ev_loop* default_loop = ev_default_loop(EVFLAG_AUTO);
+
+    struct ev_signal sigint_watcher;
+    ev_signal_init(&sigint_watcher, handle_sig, SIGINT);
+    ev_signal_start(default_loop, &sigint_watcher);
+
+    struct ev_signal sigterm_watcher;
+    ev_signal_init(&sigterm_watcher, handle_sig, SIGTERM);
+    ev_signal_start(default_loop, &sigterm_watcher);
+
+    ev_loop(default_loop, 0);
+
     return 0;
 }


src/main.c
```diff
@@ -25,10 +25,38 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+#include <signal.h>
+
+static void
+handle_sigint(
+    int sigint
+) {
+    // TODO Do proper logging here
+    printf("Caught signal SIGINT (%d)\n", sigint);
```
matthiasbeyer: Can you please change this to write on `stderr` instead?

src/main.c
```diff
@@ -25,10 +25,38 @@
...
+    exit(sigint);
+}
+
+static void
+handle_sigterm(
+    int sigterm
+) {
+    // TODO Do proper logging here
+    printf("Caught signal SIGTERM (%d)\n", sigterm);
```
matthiasbeyer: Also here, please on `stderr`.

neithernut:
Er... yeah...
The intention was to have a lib ev loop from the beginning.
Libev also supports a watcher type to listen for signals.

TheNeikos:
Okay! Gonna change to that!

neithernut:
The changes are not that excessive. You can keep the handlers but you have to adjust the definition.
Also: you can use one function for both cases (since you shut down in both cases).
Shutting down can be performed by calling `ev_break` btw.

If you need any information, do not hesitate to ask.

TheNeikos:
For now I used `EVUNLOOP_ALL` with `ev_unloop`. Should I rather use ev_break?

src/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wno-unused-parameter -pedantic")
 #
 # With a _real_ libc, we wound't have this problem.
 #
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -pthread")
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -std=gnu99 -pthread -lev")
```
neithernut: The libev linkage is done somewhere different (but was forgotten). The fix is in PR #42. I'll merge that one, you can then rebase and remove the `-lev`.

neithernut:
`ev_unloop` is acceptable.

src/CMakeLists.txt
TheNeikos: Oops! That was only a workaround.. shouldn't be in here, you are right.

TheNeikos:
Okay, this is now open for review

neithernut:
As hinted before, the signal handlers may be merged. However, the sollution is acceptable as is.
ACK.

TheNeikos:
I have merged the handlers.

neithernut:
Perfect.
ACK'd, of course.

matthiasbeyer:
Looks good to me so far. ACK.

PR #51, opened 2014-09-30
[WIP] Add array type interface

matthiasbeyer:


src/objects/array.h
```diff
@@ -28,7 +28,217 @@
...
+ * ws_array type definition
+ *
+ * The ws_array type implementation operates on `void*` internally.
+ */
+struct ws_array {
+    struct ws_object obj; //!< Supertype
+
+    struct {
+        size_t len;
+        pthread_rwlock_t rwl;
```
neithernut: Please use _one_ rw_lock for the entire array.

src/objects/array.h
```diff
@@ -28,7 +28,206 @@
...
+    struct ws_array* const self //!< Array object
+);
+
+/**
+ * Get number of used elements of an array object
+ *
+ * @return Number of used elements in the array, zero on NULL passed
+ */
+size_t
+ws_array_nused(
```
neithernut: If the array would resize automatically (double the size of the allocated space if you run out of memory), we could ommit this (and just use `len`).
Manual `resizes could also be omitted`.

src/objects/array.h
```diff
@@ -28,7 +28,206 @@
...
+ *
+ * @warning stops iteration whenever `iter` returns false
+ *
+ * @return true if the iteration exited normally (over elements was iterated),
+ * else false
+ */
+bool
+ws_array_foreach(
+    struct ws_array* const self, //!< Array object
+    bool (*iter)(void* entry, void* etc), //!< Iteration callback
```
neithernut: context comes first.

matthiasbeyer:
Discussion: I save `void*`s here. This prevents me from doing ref-counting automatically. The spec says, I should save `struct ws_object` instances, but I'd prefer this implementation.

We _could_ add a subclass for `struct ws_array`, `struct ws_object_array` for example, which does refcounting then. I guess this would be somewhat cleaner.

src/objects/array.h
```diff
@@ -28,7 +28,167 @@
...
+ *
+ * @warning stops iteration whenever `iter` returns false
+ *
+ * @return true if the iteration exited normally (over elements was iterated),
+ * else false
+ */
+bool
+ws_array_foreach(
+    struct ws_array* const self, //!< Array object
+    bool (*iter)(void* etc, void* entry), //!< Iteration callback
```
neithernut, line 159: _Maybe_ rename `etc` -> `ctx`.

neithernut:
Ok, we may want some subtypes (e.g. one for object, one or more for value types).
But for now, this is sufficient.

src/objects/array.h
matthiasbeyer, line 159: You can also pass non-context stuff, so I'd like to keep this.

matthiasbeyer:
@neithernut ACK for squashing?

neithernut:
ACK.

matthiasbeyer:
Rebased. Feel free to review again.

neithernut:
ACK.

Issue #35, opened 2014-09-29
Add string value type


neithernut:
Implement `ws_value_string`.

xVin:
I'd like to do this when I'm done with the string object type. Reading ICU documentation atm.

matthiasbeyer:
Blocked by #52 .

neithernut:
Blocks #64.

PR #52, opened 2014-09-30
Add string type

xVin:
targets #35

### DIFF
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -95,8 +95,13 @@ ws_string_init(
 struct ws_string*
 ws_string_new(void)
 {
-    //!< @todo implement
-    return NULL;
+    struct ws_string* wss = calloc(1, sizeof(*wss));
+
+    if (wss) {
+        ws_string_init(wss);
+    }
+
+    return wss;
 }
 
 size_t
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -219,7 +219,7 @@ ws_string_cmp(
     int res;
 
     ws_object_lock_read(&self->obj);
-    ws_object_lock_read(&other->obj);
+    ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
 
     res = u_strcmp(self->str, other->str);  
   
@@ -239,7 +239,7 @@ ws_string_ncmp(
     int res;
 
     ws_object_lock_read(&self->obj);
-    ws_object_lock_read(&other->obj);
+    ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
     
     res = u_strncmp(self->str + offset, other->str, n);   
 
@@ -254,8 +254,16 @@ ws_string_substr(
     struct ws_string* self,
     struct ws_string* other
 ){
-    //!< @todo implement
-    return false;
+    UChar* res;
+    ws_object_lock_read(&self->obj);
+    ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
+    
+    res = u_strstr(self->str, other->str);
+
+    ws_object_unlock(&self->obj);
+    ws_object_unlock(&other->obj);
+
+    return !!res;
 }
 
 char*
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -25,10 +25,147 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
-
 #include <stdbool.h>
 #include <stdlib.h>
 #include <string.h>
 #include <unicode/ustring.h>
 
 #include "objects/string.h"
+
+#include "objects/object.h"
+#include "objects/array.h"
+
+/*
+ *
+ *
+ * forward declarations  
+ *
+ *
+ */
+
+static bool
+deinit_callback(
+    struct ws_object* const
+);
+
+/*
+ *
+ *
+ * v-table
+ *
+ *
+ */
+
+ws_object_type_id WS_OBJECT_TYPE_ID_STRING = {
+    .supertype = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr = "ws_string",
+    
+    .init_callback = NULL,
+    .deinit_callback = deinit_callback,
+    .log_callback = NULL,
+    .run_callback = NULL,
+    .hash_callback = NULL,
+    .cmp_callback = (int (*) (struct ws_object const*, struct ws_object const*))
+                    ws_string_cmp,
+};
+
+bool
+ws_string_init(
+    struct ws_string* self
+) { 
+    //!< @todo implement
+    return false; 
+}
+
+struct ws_string*
+ws_string_new(void)
+{
+    //!< @todo implement
+    return NULL;
+}
+
+size_t
+ws_string_len(
+    struct ws_string* self
+){
+    //!< @todo implement
+    return 0;
+}
+
+struct ws_string*
+ws_string_cat(
+    struct ws_string* self,
+    struct ws_string* other
+){
+    //!< @todo implement
+    return NULL;
+}
+
+
+struct ws_string*
+ws_string_multicat(
+    struct ws_string* self,
+    struct ws_array* others
+){
+    //!< @todo implement
+    return NULL; 
+ }
+ 
+
+struct ws_string*
+ws_string_dupl(
+    struct ws_string* self
+){
+    //!< @todo implement
+    return NULL;
+}
+
+int
+ws_string_cmp(
+    struct ws_string* self,
+    struct ws_string* other
+){
+    //!< @todo implement
+    return 0;
+}
+
+int
+ws_string_ncmp(
+    struct ws_string* self,
+    struct ws_string* other,
+    size_t offset,
+    size_t n
+){
+    //!< @todo implement
+    return 0;
+}
+
+bool
+ws_string_substr(
+    struct ws_string* self,
+    struct ws_string* other
+){
+    //!< @todo implement
+    return false;
+}
+
+char*
+ws_string_raw(
+    struct ws_string* self
+){
+    //!< @todo implement
+    return NULL;   
+}
+
+/*
+ *
+ * Static Function Implementations
+ *
+ */
+static bool
+deinit_callback(
+    struct ws_object* const self
+) {
+    //!< @todo implement
+    return true;
+}
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -73,7 +73,22 @@ bool
 ws_string_init(
     struct ws_string* self
 ) { 
-    //!< @todo implement
+    if (self) {
+        self->charcount = 0; //initialize as empty string
+        self->str = calloc(self->charcount + 1, sizeof(*self->str));
+
+        if (!self->str) {
+            return false;
+        }
+
+        ws_object_init(&self->obj);
+        self->obj.id = &WS_OBJECT_TYPE_ID_STRING;
+
+        self->is_utf8 = false;
+     
+        return true;
+    }
+
     return false; 
 }
 
@@ -166,6 +181,9 @@ static bool
 deinit_callback(
     struct ws_object* const self
 ) {
-    //!< @todo implement
-    return true;
+    if (self) {
+        free(((struct ws_string*) self)->str);
+        return true;
+    }
+    return false;
 }
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -179,7 +179,7 @@ ws_string_multicat(
             ws_object_unlock(&self->obj);
             return NULL;
         } 
-   }
+    }
 
     ws_object_unlock(&self->obj);
     
@@ -191,7 +191,23 @@ struct ws_string*
 ws_string_dupl(
     struct ws_string* self
 ){
-    //!< @todo implement
+    struct ws_string* nstr = ws_string_new();
+    
+    if (nstr) {
+        ws_object_lock_read(&self->obj);
+
+        nstr->charcount = self->charcount;
+        nstr->is_utf8 = self->is_utf8;
+        nstr->str = realloc(nstr->str, (self->charcount + 1) * sizeof(*self->str));
+        nstr->str = u_strcpy(nstr->str, self->str);
+    
+        ws_object_unlock(&self->obj);
+    
+        if (nstr->str) {
+            return nstr;
+        }
+    }
+
     return NULL;
 }
 
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -124,7 +124,20 @@ ws_string_cat(
     struct ws_string* self,
     struct ws_string* other
 ){
-    //!< @todo implement
+    ws_object_lock_write(&self->obj);
+    ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
+    
+    self->str = realloc(self->str, (self->charcount + other->charcount + 1)
+                        * sizeof(*self->str));
+    self->str = u_strcat(self->str, other->str);
+
+    ws_object_unlock(&other->obj);
+    ws_object_unlock(&self->obj);
+    
+    if (self->str) {
+        return self;
+    }
+    
     return NULL;
 }
 
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -25,5 +25,10 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "objects/string.h"
 
+#include <stdbool.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unicode/ustring.h>
+
+#include "objects/string.h"
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -108,7 +108,14 @@ size_t
 ws_string_len(
     struct ws_string* self
 ){
-    //!< @todo implement
+    if (self) {
+        size_t len;
+        ws_object_lock_read(&self->obj);
+        len = self->charcount;
+        ws_object_unlock(&self->obj);
+        return len;
+    }
+
     return 0;
 }
 
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -236,8 +236,17 @@ ws_string_ncmp(
     size_t offset,
     size_t n
 ){
-    //!< @todo implement
-    return 0;
+    int res;
+
+    ws_object_lock_read(&self->obj);
+    ws_object_lock_read(&other->obj);
+    
+    res = u_strncmp(self->str + offset, other->str, n);   
+
+    ws_object_unlock(&self->obj);
+    ws_object_unlock(&self->obj);
+
+    return res;
 }
 
 bool
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -147,9 +147,44 @@ ws_string_multicat(
     struct ws_string* self,
     struct ws_array* others
 ){
-    //!< @todo implement
-    return NULL; 
- }
+    if (!self || !others) {
+        return NULL;
+    }
+
+    ws_object_lock_write(&self->obj); //!< @todo Thread-safeness!  
+    
+    int len = 0;
+    for (unsigned int i = 0; i < others->len; i++) {
+        struct ws_string* temp = ws_array_get_at(others, i);
+        if (temp && temp->str) {
+            len += temp->charcount;
+        } else {
+            ws_object_unlock(&self->obj);
+            return NULL;
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


src/objects/string.h
```diff
@@ -28,6 +28,158 @@
...
+
+/**
+* ws_string type definition
+*/
+struct ws_string {
+    struct ws_object obj; //!< Supertype
+
+    struct {
+        size_t charcount;
+        pthread_rwlock_t rwl;
```
neithernut: Same as for the array type by @matthiasbeyer: one lock is sufficient for the entire type.

src/objects/string.h
```diff
@@ -28,6 +28,158 @@
...
+    } charcount; //!< Atomic charcount member
+
+    struct {
+        char* firstbyte;
+        pthread_rwlock_t rwl;
+    } firstbyte; //!< Atomic member for first byte
+
+    struct {
+        char* lastbyte;
```
neithernut: Can be derived from firstbyte and bytecount.

src/objects/string.c
```diff
@@ -25,5 +25,140 @@
...
+ *
+ *
+ * forward declarations  
+ *
+ *
+ */
+
+static bool
+deinit_callback(
+struct ws_object* const
```
matthiasbeyer: Indention missing.

src/objects/string.c
```diff
@@ -25,5 +25,140 @@
...
+    struct ws_string* self
+){
+    //!< @todo implement
+    return NULL;   
+}
+
+/* Static Function Implementations*/
+static bool
+deinit_callback(
+struct ws_object* const self
```
matthiasbeyer: Missing indention.

src/objects/string.c
```diff
@@ -25,5 +25,209 @@
...
+){
+    ws_object_lock_write(&self->obj);
+    ws_object_lock_read(&other->obj);
+
+    self->str = u_strcat(self->str, const other->str);
+
+    ws_object_unlock_read(&other->obj);
+    ws_object_unlock_write(&self->obj);
+    
+    if (self) {
```
matthiasbeyer: The condition should be `self->str`, shouldn't it?

src/objects/string.c
```diff
@@ -25,5 +25,209 @@
...
+
+struct ws_string*
+ws_string_multicat(
+    struct ws_string* self,
+    struct ws_array* others
+){
+    ws_object_lock_write(&self->obj);   
+
+    for (int i = 0; i < others->len; i++) {
+        self = ws_string_cat(self, others[i]);
```
matthiasbeyer: You do not check the return value of `ws_string_cat()` here!

matthiasbeyer: `others[i]` doesn't work. The array type has a function for getting an element at a specific index!

src/objects/string.c
```diff
@@ -25,5 +25,209 @@
...
+    
+    struct ws_string* nstr = ws_string_new();
+
+    nstr->charcount = self->charcount;
+    nstr->is_utf8 = self->is_utf8;
+    nstr->str = u_strcpy(nstr->str, self->str);
+    
+    ws_object_unlock_read(&self->obj);
+    
+    if (nstr) {
```
matthiasbeyer: The condition should be `nstr->str`, shouldn't it?

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+    .deinit_callback = deinit_callback,
+};
+
+bool
+ws_string_init(
+    struct ws_string* self
+) { 
+    if (self) {
+        self->charcount = 0; //initialize as empty string
+        self->str = calloc(self->charcount+1, sizeof(*self->str));
```
matthiasbeyer: Spaces around `+` missing.

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+
+        if (!self->str) {
+            return false;
+        }
+
+        memset(self->str, '\0', sizeof(*self->str));
```
matthiasbeyer: This does not work. You have to `memset` the actual allocated space here, so you should pass `self->charcount + 1` here!

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+ 
+    return NULL; 
+ }
+ 
+
+struct ws_string*
+ws_string_dupl(
+    struct ws_string* self
+){
+    ws_object_lock_read(&self->obj);
```
matthiasbeyer: You can move this line to _after: the initialization of the new string (after the `ws_string_new()` call), to save time here.

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+    
+    struct ws_string* nstr = ws_string_new();
```
matthiasbeyer, line 194: You do not check the return value, it _could_ be `NULL`

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+    struct ws_string* other,
+    size_t offset,
+    size_t n
+){
+    int res;
+
+    ws_object_lock_read(&self->obj);
+    ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
+    
+    res = u_strcmp(self->str+offset, other->str);
```
matthiasbeyer: Nope, `ncmp` should compare the first `n` characters. There must be something like `u_strncmp` or something.

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+bool
+ws_string_substr(
+    struct ws_string* self,
+    struct ws_string* other
+){
+    if (ws_string_cmp(self, other) == 0) {
+        return true;
+    }
+
+    else {
```
matthiasbeyer: The else should be on the line of the closing if bracket, so `} else {` should be on the appropriate line.

Anyways, I assume that the implementation of this is not ready yet.

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+    else {  
+        return false;
+    }
+}
+
+char*
+ws_string_raw(
+    struct ws_string* self
+){
+    if (!self->is_utf8) {
```
matthiasbeyer, line 275: This is not read-locked.

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+        return NULL;
+    }
+
+    char* output = calloc(self->charcount, sizeof(char));
+    int32_t destL;
+    UErrorCode err;    
+
+    ws_object_lock_read(&self->obj);
+
+    output = u_strToUTF8(output, self->charcount, &destL, self->str,
```
matthiasbeyer: What does the `u_strToUTF8()` call?

src/objects/string.c
```diff
@@ -25,5 +25,252 @@
...
+char*
+ws_string_raw(
+    struct ws_string* self
+){
+    if (!self->is_utf8) {
+        return NULL;
+    }
+
+    char* output = calloc(self->charcount, sizeof(char));
+    int32_t destL;
```
matthiasbeyer: Please don't use CamelCase but snake_case. `dest_len` or just `len` should be enough here!

matthiasbeyer:
Feel free to do a squash, to clean the history. It's cleaner for you to work with, afterwards.

src/objects/string.c
xVin: Gonna fix this and put everything up for review again!

src/objects/string.c
neithernut: Wait...
is the `memset` even neccessary? She used `calloc`...

src/objects/string.c
neithernut: It transforms the internal representation to an utf string and puts it into output. Should be obvious from the name.

matthiasbeyer:
@xVin Have a look at [this](https://github.com/RaphaelJ/fast_strstr), it _could_ be usefull for the `ws_string_substr()` implementation.

I don't know, but guess, that BSD3C license is compatible with our license, @neithernut , am I right? We could use this code for speeding up some of our string operations (namingly `ws_string_substr()`). We would then do a copy-paste, I assume.

matthiasbeyer:
@neithernut , @xVin [this page](https://www.gnu.org/licenses/license-list.html#GPLCompatibleLicenses) says that BSD-3-clause is compatible with GPL.

xVin:
I'll have a deeper look at the fast_strstr lib later, but I dunno if it's worth the extra effort - we'd have to convert it to UChars somehow.
For now, I just used ICU's strstr function.

I cleaned up my history (with a little help from my friends), added ws_string_substring and removed the memset in ws_string_init().
Ready for review.

matthiasbeyer:
I'm not sure about how much effort this would be, but I guess it wouldn't be that much, as you literally only compare memory. So I guess there is no need to adapt it!

But I'm not that sure about it and would like to hear @neithernut s opinion on this.

I'll review later, too tired right now, so this has to wait until tomorrow!

neithernut:
As long as we don't need substrings we should not put in the extra effort.

matthiasbeyer:
Please rebase onto the latest `develop` and remove all  `ws_object_unlock_read()` and `ws_object_unlock_write()` calls in favour of the new `ws_object_unlock()` call.

Feel free to do the fixups directly, you don't have to push the single "Fixup: ..." commits!

src/objects/string.c
```diff
@@ -25,5 +25,259 @@
...
+
+struct ws_string*
+ws_string_cat(
+    struct ws_string* self,
+    struct ws_string* other
+){
+    ws_object_lock_write(&self->obj);
+    ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
+
+    self->str = u_strcat(self->str, other->str);
```
neithernut, line 132: You need to reallocate `self->str`, so the resulting string is fitting in the memory allocated.

src/objects/string.c
```diff
@@ -25,5 +25,259 @@
...
+
+struct ws_string*
+ws_string_multicat(
+    struct ws_string* self,
+    struct ws_array* others
+){
+    ws_object_lock_write(&self->obj); //!< @todo Thread-safeness!  
+
+    for (unsigned int i = 0; i < others->len; i++) {
+        self->str = u_strcat(self->str, ws_array_get_at(others, i));
```
neithernut: Same here: you need to reallocate `self->str` beforehand.

matthiasbeyer: And because a `realloc()` call is expensive it would be good to calculate the new length beforehand and not doing a `realloc()` call in every iteration!

xVin: I figured that myself :)

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+    if (self) {
+        self->charcount = 0; //initialize as empty string
+        self->str = calloc(self->charcount + 1, sizeof(*self->str));
+
+        if (!self->str) {
+            return false;
+        }
+
+        ws_object_init(&self->obj);
+        self->obj.id = &WS_OBJECT_TYPE_ID_ARRAY;
```
matthiasbeyer: This is actually not true, you have the string type here, don't you?

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+        return true;
+    }
+
+    return false; 
+}
+
+struct ws_string*
+ws_string_new(void)
+{
+    struct ws_string* wss = calloc(1, sizeof(struct ws_string));
```
matthiasbeyer: A shortcut would be (for the allocating call) `calloc(1, sizeof(*wss))`

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+struct ws_string*
+ws_string_multicat(
+    struct ws_string* self,
+    struct ws_array* others
+){
+    ws_object_lock_write(&self->obj); //!< @todo Thread-safeness!  
+    
+    int len = 0;
+    for (unsigned int i = 0; i < others->len; i++) {
+        if (((struct ws_string*) ws_array_get_at(others, i))->str) {
```
matthiasbeyer: That's not good. If the `ws_array_get_at()` call returns NULL, you segfault here!

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+            len += ((struct ws_string*) ws_array_get_at(others, i))->charcount;
```
matthiasbeyer: Same as for the line above.

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+        } else {
+            return NULL;
+        }
+    }
+
+    self->str = realloc(self->str, (len +1) * sizeof(*self->str));
+
+    for (unsigned int i = 0; i < others->len; i++) {
+        self->str = u_strcat(self->str,
+                            ((struct ws_string*)ws_array_get_at(others, i))->str);
```
matthiasbeyer, line 176: Here too, this could cause segfaults.

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+ws_string_raw(
+    struct ws_string* self
+){
+    ws_object_lock_read(&self->obj);
+
+    if (!self->is_utf8) {
+        return NULL;
+    }
+
+    char* output = calloc(self->charcount, sizeof(char));
```
matthiasbeyer: `sizeof(*output)` - it's the same, but its more maintainable.

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+/*
+ *
+ * Static Function Implementations
+ *
+ */
+static bool
+deinit_callback(
+    struct ws_object* const self
+) {
+    free(((struct ws_string*) self)->str);
```
matthiasbeyer: If `self` is `NULL`, you will also segfault here.

xVin: Ok, I modeled this after the deinit() in ws_array, should be fixed there, too, then?

matthiasbeyer: sure.

src/objects/string.c
```diff
@@ -25,5 +25,273 @@
...
+        if (!self) {
+            ws_object_unlock(&self->obj);
+            return NULL;
+        } 
+    }
+
+    ws_object_unlock(&self->obj);
+    
+    if (self) {
+        return self;
```
matthiasbeyer: You don't need an `if` here, you can just return `self`.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <stdbool.h>
+#include <unicode/ustring.h>
+#include <string.h>
```
matthiasbeyer: Sort them please.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+ * v-table
+ *
+ *
+ */
+
+ws_object_type_id WS_OBJECT_TYPE_ID_STRING = {
+    .supertype = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr = "ws_string",
+    .deinit_callback = deinit_callback,
+};
```
matthiasbeyer, line 70: Please add the other entries of the struct and set them to `NULL`.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+    return NULL;
+}
+
+
+struct ws_string*
+ws_string_multicat(
+    struct ws_string* self,
+    struct ws_array* others
+){
+    if (!self) {
```
matthiasbeyer: `!self || !others` <- how about that? :-)

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+        return NULL;
+    }
+
+    ws_object_lock_write(&self->obj); //!< @todo Thread-safeness!  
+
```
matthiasbeyer, line 155: You maybe should lock the array object here, too.

But I guess you have to ask @neithernut for this.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+        return NULL;
+    }
+
+    ws_object_lock_write(&self->obj); //!< @todo Thread-safeness!  
+    
+    int len = 0;
+    for (unsigned int i = 0; i < others->len; i++) {
+        struct ws_string* temp = ws_array_get_at(others, i);
+        if (temp && temp->str) {
+                len += temp->charcount;
```
matthiasbeyer: Too much indentation here.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+        } else {
+            ws_object_unlock(&self->obj);
+            return NULL;
+        }
+    }
+
+    self->str = realloc(self->str, (len + 1) * sizeof(*self->str));
+
+/* This loop doesn't check for NULL in others, assuming that in the final
```
matthiasbeyer: ... in `others` ... (add the backticks around the "others", to make it a code fragment in the documentation.

src/objects/string.c
xVin, line 155: Yeah we should. Itentionally not implemented yet b/c locking solution not ready yet.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+    res = u_strstr(self->str, other->str);
+
+    ws_object_unlock(&self->obj);
+    ws_object_unlock(&other->obj);
+
+    if (res) {
+        return true;
+    } else {  
+        return false;
+    }
```
matthiasbeyer: Can be shortened to `return !!res;`

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+ws_string_raw(
+    struct ws_string* self
+){
+    ws_object_lock_read(&self->obj);
+
+    if (!self->is_utf8) {
+        return NULL;
+    }
+
+    char* output = calloc(self->charcount, sizeof(*output));
```
matthiasbeyer: Not `self->charcount + 1` (plus one for the trailing zero-byte)?

matthiasbeyer:
Gets better with every update!

Btw.: The others ( @neithernut , @TheNeikos , @manuelmessner ) should review too!

xVin:
Done with fixing stuff.

matthiasbeyer:
@neithernut can you review, too?

matthiasbeyer:
@xVin You need to rebase on the latest `develop`, merge conflicts will occur.

src/objects/string.c
```diff
@@ -25,5 +25,284 @@
...
+ws_object_type_id WS_OBJECT_TYPE_ID_STRING = {
+    .supertype = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr = "ws_string",
+    
+    .init_callback = NULL,
+    .deinit_callback = deinit_callback,
+    .log_callback = NULL,
+    .run_callback = NULL,
+    .hash_callback = NULL,
+};
```
neithernut, line 70: We added a compare-callback lately. Would be nice if you could add an implementation. You may simply pass the `ws_string_cmp` function, but you will probably have to cast the function type.

xVin, line 70: Just to clarify: Does that mean declaring ws_string_cmp as a static function (like deinit)?

neithernut, line 70: No, it can stay a "public" function, at least in my opinion.

xVin, line 70: Okay, but I'm not really sure how to cast the function type then.

neithernut:
Ok, looks good. ACK from me.

src/objects/string.c
```diff
@@ -25,5 +25,286 @@
...
+ws_object_type_id WS_OBJECT_TYPE_ID_STRING = {
+    .supertype = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr = "ws_string",
+    
+    .init_callback = NULL,
+    .deinit_callback = deinit_callback,
+    .log_callback = NULL,
+    .run_callback = NULL,
+    .hash_callback = NULL,
+    .cmp_callback = (int (*) (struct ws_object const*, struct ws_object const*))
```
matthiasbeyer, line 68: Here shouldn't be a cast. The cast should be in the cmp() method. I would prefer a static function which simply does the call to `ws_string_cmp()` over this cast!

If you change the interface definition of the `cmp_callback`, this causes the code to compile anyways, but to crash if the callback gets called while running.

neithernut, line 68: This is unneccessary. The cast will fail if the definition changes!

xVin, line 68: I agree with neithernut.

matthiasbeyer:
Okay, the latest comment was my fault. So ACK from me.

neithernut:
Happy squashing!

matthiasbeyer:
No squashing required here, at least in my opinion.

neithernut:
k, then ACK.

matthiasbeyer:
@neithernut go ahead and merge! :-P

PR #56, opened 2014-09-30
[WIP] Add logger module

matthiasbeyer:

PR #61, opened 2014-09-30
Fix: Redefinition of __ws_alloc_size__ by renaming

matthiasbeyer:
Redifinition "overloading" like is not allowed in C. Not even in macros.

Blocks #56


neithernut:
ACK.

PR #79, opened 2014-10-01
Fix logger

neithernut:
This PR fixes a few issues with the logger.
It also initializes the logger in the mian function.


matthiasbeyer:
ACK.

PR #80, opened 2014-10-01
ws_log() reimplementation for newline character appending

matthiasbeyer:
Please review carefully.

### DIFF
--- a/src/logger/module.c
+++ b/src/logger/module.c
@@ -84,30 +84,35 @@ ws_log(
     char* fmt,
     ...
 ) {
-    char* __fmt = fmt;
     va_list list;
-    va_start(list, fmt);
+    size_t pref_len = 0;
+    size_t buf_len = strlen(fmt);
 
-    bool used_prefix = false;
-
-    if (ctx) {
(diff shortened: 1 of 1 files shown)


src/logger/module.c
```diff
@@ -84,30 +84,36 @@ ws_log(
     char* fmt,
     ...
 ) {
-    char* __fmt = fmt;
     va_list list;
-    va_start(list, fmt);
+    size_t fmt_offset = 0;
+    size_t buf_len = 0;
+    size_t pref_len = 0;
```
neithernut, line 88: This can and should be moved to line 95 (into the `if`). Or, optionall, remove `fmt_offset`. Having both is unneccessary.

matthiasbeyer, line 88: Nope. It is used in another scope as well. (see line  106)

neithernut, line 88: Yes, but you can use `pred_len` there, too.
In your implementation, _both_ vars have the same value. It doesn't make sense to have both of them.

neithernut:
ACK.

Issue #65, opened 2014-09-30
Implement: ws_set type


matthiasbeyer:
Implement the set type for the specification.

The hashtable type in the class hierarchy can be emitted, but some methods might be transferred to the set type.

neithernut:
Blocks #64.

matthiasbeyer:
Blocks #72

neithernut:
Blocks #114.

PR #86, opened 2014-10-01
[WIP] Add set type

matthiasbeyer:
This PR is meant to implement the set type and therefor solve #65 .

### DIFF
--- a/src/objects/set.c
+++ b/src/objects/set.c
@@ -229,6 +229,10 @@ size_t
 ws_set_cardinality(
     struct ws_set const* self
 ) {
+    if (self) {
+        return r_set_cardinality(self->set);
+    }
+
     return 0;
 }
 
--- a/src/objects/set.c
+++ b/src/objects/set.c
@@ -25,6 +25,7 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
 #include <stdlib.h>
 
 #include <libreset/hash.h>
@@ -33,4 +34,102 @@
 #include "objects/object.h"
 
 #include "objects/set.h"
+
+int
+ws_set_init(
+    struct ws_set* self
+) {
+    return;
+}
+
+struct ws_set*
+ws_set_new(void)
+{
+    return NULL;
+}
+
+int
+ws_set_insert(
+    struct ws_set* self,
+    struct ws_object* obj
+) {
+    return 0;
+}
+
+int
+ws_set_remove(
+    struct ws_set* self,
+    struct ws_object const* cmp
+) {
+    return 0;
+}
+
+struct ws_object*
+ws_set_get(
+    struct ws_set const* self,
+    struct ws_object const* cmp
+) {
+    return NULL;
+}
+
+int
+ws_set_union(
+    struct ws_set* dest,
+    struct ws_set const* src_a,
+    struct ws_set const* src_b
+) {
+    return 0;
+}
+
+int
+ws_set_intersection(
+    struct ws_set* dest,
+    struct ws_set const* src_a,
+    struct ws_set const* src_b
+) {
+    return 0;
+}
+
+int
+ws_set_xor(
+    struct ws_set* dest,
+    struct ws_set const* src_a,
+    struct ws_set const* src_b
+) {
+    return 0;
+}
+
+bool
+ws_set_is_subset(
+    struct ws_set const* self,
+    struct ws_set const* other
+) {
+    return false;
+}
+
+bool
+ws_set_equal(
+    struct ws_set const* self,
+    struct ws_set const* other
+) {
+    return false;
+}
+
+size_t
+ws_set_cardinality(
+    struct ws_set const* self
+) {
+    return 0;
+}
+
+int
+ws_set_select(
+    struct ws_set const* self,
+    ws_set_predf pred,
+    void* pred_etc,
+    ws_set_procf proc,
+    void* proc_etc
+) {
+    return 0;
+}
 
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
It seems that the libreset dependencies are missing. @neithernut Aren't they already added? :question:

neithernut:
The functions popping up are functions which are not yet implemented.
Of course you get linker errors if you try to use them now.
Just comment them out and insert some stub code until libreset ships them.

matthiasbeyer:
Note: The commit which outcommented the functions which use the libreset functionality which is not yet shipped can be `git revert`ed very simply lateron!

src/objects/set.c
```diff
@@ -0,0 +1,228 @@
...
+) {
+    if (self) {
+        ws_object_init(&self->obj);
+
+        self->obj.id = &WS_OBJECT_TYPE_ID_SET;
+
+        self->set_config.cmpf = NULL;
+        self->set_config.copyf = NULL;
+        self->set_config.freef = NULL;
+        self->set_config.hashf = NULL;
```
neithernut: We acutally _have_ methods to perform some of the tasks, so we don't have to set them to `NULL`:
- `copyf` can be set to `ws_object_getref()`
- `freef` can be set to `ws_object_unref()`
- `hashf` can be set to `ws_object_hash()`

src/objects/set.h
```diff
@@ -0,0 +1,226 @@
...
+ *
+ * The ws_set type implementation operates on libreset internally.
+ * The `struct ws_set` type is only used for `ws_object` objects, so nobody has
+ * to care about the set configuration. Therefor, this is hold internally to the
+ * set interface.
+ */
+struct ws_set {
+    struct ws_object obj; //!< superclass
+
+    struct r_set_cfg set_config; //!< Configuration for libreset sets
```
neithernut: The config is a global constant.

src/objects/set.c
matthiasbeyer: Are you sure? The hash, okay... but the others? The callbacks are meant to actually _copy_ or _free_ something!

src/objects/set.h
matthiasbeyer: You're sure about this? Are you sure we never ever need set functionality _without_ the internal object types?

neithernut: Yes. Even if we ever need a `ws_set` for non-objects, we can wrap the thing in an object and probably should have a proper object type for the thing anyways.
If you look at the interface, you'll notice that the set object type is meant to be used only for objects.

matthiasbeyer:
Targets #65 btw.

src/objects/set.c
```diff
@@ -0,0 +1,277 @@
...
+ *
+ * Interface implementation
+ *
+ */
+
+int
+ws_set_init(
+    struct ws_set* self
+) {
+    if (self) {
```
neithernut: Would prefer an early return.

src/objects/set.c
```diff
@@ -0,0 +1,277 @@
...
+
+    return set;
+}
+
+int
+ws_set_insert(
+    struct ws_set* self,
+    struct ws_object* obj
+) {
+    if (self && obj) {
```
neithernut: Would prefer an early return.

src/objects/set.h
```diff
@@ -0,0 +1,225 @@
...
+ * Remove an object from the set by a unique identification
+ *
+ * @memberof ws_set
+ *
+ * @return zero on success else negative error value from errno.h
+ */
+int
+ws_set_remove(
+    struct ws_set* self //!< The set
+    /* object-UUID */
```
neithernut: Why is this not implemented yet? All you need is pass an `ws_object`.

src/objects/set.h
```diff
@@ -0,0 +1,225 @@
...
+ * Get an object from a set
+ *
+ * @memberof ws_set
+ *
+ * @return the ws_object object or NULL on failure
+ */
+struct ws_object*
+ws_set_get(
+    struct ws_set const* self //!< The set
+    /* object-UUID */
```
neithernut: Same here.

src/objects/set.h
matthiasbeyer: No, because we have (or should have) a unique id for objects, so it would be much better to just pass the UID of an object around.

src/objects/set.h
matthiasbeyer: See above.

src/objects/set.c
```diff
@@ -0,0 +1,277 @@
...
+ *
+ */
+
+static int
+cmp_objects(
+    void const* a,
+    void const* b
+) {
+    // comparing the pointers is sufficient --at least for now.
+    return a == b;
```
neithernut, line 282: We should add a TODO here (needs code in `ws_object`)

src/objects/set.h
neithernut: No, we do _not_ have a UUID.

matthiasbeyer:
I guess, besides the functionality which is not yet shipped by libreset, this is ready.

src/objects/set.c
```diff
@@ -0,0 +1,290 @@
...
+
+/**
+ * Type information for ws_set type
+ */
+ws_object_type_id WS_OBJECT_TYPE_ID_SET = {
+    .supertype  = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr    = "ws_set",
+
+    .deinit_callback = deinit_set,
+};
```
neithernut, line 92: You need to `NULL` the callbacks you do not set.

neithernut:
GO for squashing.

neithernut:
ACK.

PR #88, opened 2014-10-01
Move: ws_array_init() to init callback

matthiasbeyer:
The ws_object type allows init callbacks, so the init function for the
object type should be used here.

One addition is added as well: The init callback does a type check
internally and returns false if the type is not an array.


neithernut:
Seems to be ok.
ACK.

Issue #64, opened 2014-09-30
Add callback for access of object internas


neithernut:
We will need some way to implement an index operator for our transaction. This should be done by adding a callback to the object type.
It should take a `ws_string` and set/return a `ws_value`.
If that callback is `NULL`, the function should return `nil` automatically.

matthiasbeyer:
An _index_ operator for the _object_ type?

neithernut:
Yep. So you cann access internal values of an object, like the path of the background image for a monitor.

matthiasbeyer:
Shouldn't such functionality be covered by the `set` type?

neithernut:
Nope, that way wou would need to put a set in each and every object, just to access e.g. the width and the height of a window.
That would be completely and utterly unneccessary.

matthiasbeyer:
Ah, ok. You mean something like a key-value storage, where you can access the values by key, right? And this is required for the API, right?

So... this is covered by the `ws_value_named` type, isn't it? We just need a way to put multiple of this (multiple key->value accociations) into one object.

neithernut:
This would introduce imense overhead.

neithernut:
@matthiasbeyer and I came up with a nice solution:

Each object type will have a sorted table with the following entries:
- Member name
- Member type
- Member offset in the real object
- Access modifiers (read/write)

The indexing method can then be implemented in the object type. Both the get and the set method will take a string and a value as a parameter. As the value denotes its type, the function can now search for a matching entry in the table and transfer/convert the internal datatype into the value apssed to the function.
Storing a value works similarly.

PR #91, opened 2014-10-02
Add object attribute mechanism

matthiasbeyer:
This _kindof_ sucks.

This is my first idea of how to implement object attribute getters for
the object type.

Each type must provide an object-name-offset-type table, which gets
stored in the object v-table. The function `ws_object_attr_read()` allows
the _reading_ of such an attribute.

This code is untested by now and I consider this code as really ugly,
but I don't know a better way to do it by now. Please stay calm when
reviewing this.

Targets #64


matthiasbeyer:
I don't understand why the last build fails. It seems that the code is not allowed in C, so I guess I have to remove 64c9a8e, what do you guys think? @neithernut ? @TheNeikos ?

neithernut:
We use `-pedantic`, which does not allow such "dirty" constructs.
Use a basepointer instead and put the attribute implementation in another static constant.

matthiasbeyer:
@neithernut Review again please and tell me what to do.

src/objects/object.h
```diff
@@ -97,6 +97,35 @@ typedef size_t (*ws_object_hash_callback)(struct ws_object* const);
  */
 
 /**
+ * Attribute type identifier
+ */
+enum ws_object_attribute_type {
+    WS_OBJ_ATTR_NO_TYPE         = 0,
+    WS_OBJ_ATTR_PLAIN           = 1 << 0,
```
neithernut: Is there a reason you made them single bits?

src/objects/object.h
```diff
@@ -357,6 +388,33 @@ ws_object_deinit(
...
+ *
+ * @note Does string-match the attribute name with the `ident` parameter in O(n)
+ *
+ * @return zero on success, else negative error code from errno.h
+ */
+int
+ws_object_attr_read(
+    struct ws_object* self, //!< The object
+    char* ident, //!< The identifier for the attribute
+    enum ws_object_attribute_type* type_dest, //!< Dest. of the type information
```
neithernut: Instead of `type_dest` and `dest` of a type identifier and void**, we simply should pass a value.

src/objects/object.h
matthiasbeyer: Yes. You could have an `WS_OBJ_ATTR_TYPE_OBJ | WS_OBJ_ATTR_PTR`, for example.

src/objects/object.h
matthiasbeyer: And then cast the `ws_value` type to the appropriate type doing a 'realloc()'?

src/objects/object.h
neithernut: The referred variable's is of exactely one type.
Also, this makes your usage of arrays (for sizes) completely nonsensical.

src/objects/object.h
neithernut: Nope, you read the type and either fail or set the value to the type read out.
You obviously might include the conversion functions in the table you now use for the size info.

src/objects/object.h
matthiasbeyer: So you think I should add a `WS_..._INT_PTR` alongside with `WS_..._INT` ?

src/objects/object.h
matthiasbeyer: So you mean something like "failed because types do not match", when talking about failiing here? Would be an option, yes.

src/objects/object.h
neithernut: As the intention is to set _attributes_, you should not have poitner variants at all.
Having pointers to single basic values (for something different than arrays) it bad design.

src/objects/object.h
neithernut: Yep.

src/objects/object.h
matthiasbeyer: But not in case of the object type. You actually could have a `ws_set` as attribute for a monitor, for example. I don't know if we want to embedd this attribute into the actual monitor object, do we?

neithernut: We do, else we probably have to call a _function_ to get the value, anyway.

matthiasbeyer: I'll do a fixup right now.

matthiasbeyer:
@neithernut can you review and ack my new interface def, so I can start reimplement the function?

src/objects/object.h
```diff
@@ -99,6 +100,33 @@ typedef size_t (*ws_object_hash_callback)(struct ws_object* const);
...
 
 /**
+ * Attribute type identifier
+ */
+enum ws_object_attribute_type {
+    WS_OBJ_ATTR_NO_TYPE = 0,
+
+    WS_OBJ_ATTR_TYPE_CHAR,
+    WS_OBJ_ATTR_TYPE_INT,
+    WS_OBJ_ATTR_TYPE_UINT,
```
neithernut: Please specify the size here, like `UINT32`.

neithernut:
I would like to see callbacks in the type-table, which can convert from the basic type to the value representation and back. You may omit the size for them.

matthiasbeyer:
Waiting for the actual implementations of the value types.

matthiasbeyer:
Note: I added a convenience mechanism for getting the appropriate `ws_value` subtype as an instance for a member. This is a fancy hack, as the whole attribute thingy is, as it is (kindof) meta-programming in C. 

Don't know whether such an functionality is wanted, but I consider it as useful. Please review.

matthiasbeyer:
@TheNeikos Would like you to review this. Also notice my top comment. I still think this is a load of shit (what we do here), but it is implemented as clean as I can, so I consider it as merge-ready somehow.

@manuelmessner , @neithernut , @xVin would like to hear your opinions as well!

src/objects/object.c
```diff
@@ -38,11 +47,75 @@
...
+ * Attribute information about the type
+ */
+struct ws_object_attribute const WS_OBJECT_ATTRS_OBJECT[] = {
+/* The following code is outcommented as it should be an example only: */
+//    {
+//        .name = "id",
+//        .offset_in_struct = offsetof(struct ws_object, id),// stddef: offsetof
+//        .type = WS_OBJ_ATTR_NO_TYPE,
+//    }
+    { NULL, 0, 0 },
```
TheNeikos: I think `.name` is NULL here? Could you write the full names?

src/objects/object.c
```diff
@@ -38,11 +47,75 @@
...
+    [WS_OBJ_ATTR_TYPE_UINT64]   = {
+        .type = WS_VALUE_TYPE_INT,
+        .size = sizeof(struct ws_value_int),
+        .init = (void (*)(struct ws_value*)) ws_value_int_init,
+    },
+
+    [WS_OBJ_ATTR_TYPE_DOUBLE]   = {
+        .type = WS_VALUE_TYPE_INT,
+        .size = sizeof(struct ws_value_int),
+        .init = (void (*)(struct ws_value*)) ws_value_int_init,
```
TheNeikos: Should this be `int_init`? (Even though it's a double)

src/objects/object.c
```diff
@@ -303,6 +376,145 @@ ws_object_deinit(
...
+    char* ident,
+    struct ws_value* dest
+) {
+    if (!self || !ident || !dest) {
+        return -EINVAL;
+    }
+
+    ws_object_lock_read(self);
+
+    if (!self->id || !self->id->attribute_table) {
```
TheNeikos: Could this be moved above the lock? Or wouldn't it be threadsafe?

src/objects/object.c
```diff
@@ -303,6 +376,145 @@ ws_object_deinit(
...
+        ws_object_unlock(self);
+        return -EINVAL;
+    }
+
+    bool found;
```
TheNeikos: Could this be set to true by accident as it is stack alloced?

src/objects/object.c
matthiasbeyer: There is no `ws_value_double` type yet, so I use the int type here, for now. We should discuss if we even need a double type...

src/objects/object.c
matthiasbeyer: No, it must be after the lock. I'm reading members of the object here, so I need the lock beforehand.

src/objects/object.c
matthiasbeyer: Don't know, but I'll add a init value for this one.

src/objects/object.c
neithernut: Well...
The id and it's members are practical constants, so it _is_ save to read them before locking.

neithernut:
You can progress here, now that all the deps are cleared.

matthiasbeyer:
Not really, the `ws_value_set` type is still missing.

matthiasbeyer:
String type implementation added, waiting on `ws_value_set` type.

matthiasbeyer:
I consider this as "ready" now, as tests pass. Please review carefully.

Missing features:
- String implemation
- Char implementation
- Should work for int32/64, but does not differ between them (no value types)
- No diff between signed/unsigned ints (no value types)
- Doubles (no value types)
- object-reference implementation
- Named-value implementation
- boolean implementation

So, what is included? Literally the implementation for `int`.

But the other stuff can be added easily, if one knows the infrastructure!

src/objects/object.c
```diff
@@ -55,6 +82,53 @@ ws_object_type_id WS_OBJECT_TYPE_ID_OBJECT = {
     .run_callback = NULL,
     .hash_callback = NULL,
     .cmp_callback = NULL,
+    .attribute_table = WS_OBJECT_ATTRS_OBJECT,
+};
+
+static const struct {
+    enum ws_value_type type;
+    size_t size;
```
neithernut: Do you actually need the size here? If not, remove it.

src/objects/object.c
```diff
@@ -335,6 +409,132 @@ ws_object_deinit(
 }
 
 int
+ws_object_attr_read(
+    struct ws_object* self,
+    char* ident,
+    struct ws_value* dest
+) {
+    if (unlikely(!self || !ident || !dest || !self->id ||
```
neithernut: Would prefer `__ws_nonnull__` for self and dest.

src/objects/object.c
```diff
@@ -335,6 +409,132 @@ ws_object_deinit(
...
+
+
+    ws_object_lock_read(self);
+
+    bool found = false;
+    size_t offset = 0;
+    enum ws_object_attribute_type type = WS_OBJ_ATTR_NO_TYPE;
+
+    struct ws_object_attribute const* iter = &self->id->attribute_table[0];
+    for (; iter->name && !found; iter++) {
```
neithernut: This is kinda ugly. You declare the iterator _before_ the `for`, not in it. However, you still have a `for` instead of using a `while`.

src/objects/object.c
```diff
@@ -335,6 +409,132 @@ ws_object_deinit(
...
+}
+
+enum ws_object_attribute_type
+ws_object_attr_type(
+    struct ws_object* self,
+    char* ident
+) {
+    ws_object_lock_read(self);
+
+    if (!self || !self->id || !self->id->attribute_table) {
```
neithernut: Would prefer `__ws_nonnull__`.

src/objects/object.c
```diff
@@ -335,6 +409,132 @@ ws_object_deinit(
...
+    ws_object_unlock(self);
+    return 0;
+}
+
+enum ws_object_attribute_type
+ws_object_attr_type(
+    struct ws_object* self,
+    char* ident
+) {
+    ws_object_lock_read(self);
```
neithernut, line 513: `self` may be `NULL`, which wouldn't be a problem if you ued `__ws_nonnull__`.

matthiasbeyer, line 513: It may even be NULL if I used `__ws_nonnull__`, we had that discussion before. Anyways: Fixed.

matthiasbeyer:
And now we have the perfect case: The function which causes the segfault gets `NULL` as parameter while it is tagged with `__ws_nonnull__`. The compiler does not care about the tag and still builds it and when you run it, there is no runtime test for this. But hey, I was told to remove the runtime checks... :angry:

matthiasbeyer:
Revert-Fixes added, will clean history as soon as the bitchfight here is over.

matthiasbeyer:
Fuck, there is no "pissed"-smilie available.

neithernut:
Is this still WIP? Note that we could add string, set and other stuff sometime later. (I'd like something to fiddle around with)

matthiasbeyer:
This is ready, but I would like you to review _really_ carefully. I'd even suggest a pair-review or something like that with @TheNeikos or so.

neithernut:
Ok, maybe this evening.

matthiasbeyer:
Updated for `struct ws_object*` attributes.

src/objects/object.c
```diff
@@ -57,6 +84,44 @@ ws_object_type_id WS_OBJECT_TYPE_ID_OBJECT = {
...
+        .type = WS_VALUE_TYPE_INT,
+    }, //!< @todo double?
+
+    [WS_OBJ_ATTR_TYPE_STRING]   = {
+        .type = WS_VALUE_TYPE_STRING,
+    },
+
+//    [WS_OBJ_ATTR_TYPE_OBJ]      = {
+//        .type = WS_VALUE_TYPE_OBJECT_ID,
+//    },
```
neithernut: You forgot about this bit.

neithernut:
What's the status on this?

matthiasbeyer:
I will do a complete check for this today and write up whats missing.

matthiasbeyer:
Q: Shall we differ between `u/int32/64` ? They are all mapped on `intmax_t` anyways!

matthiasbeyer:
I consider this as ready. The string functionality (reading a `char*` and getting a `ws_string` from it) does not work as the `ws_string` implementation is still buggy. But it can be implemented really simple and tests are also prepared for it, so one has only to revert the "outcommented" commit.

neithernut:
Of course we should differentiate, as you don't use `intmax_t` all of the the structs!

matthiasbeyer:
Ok.

neithernut:
ACK.

matthiasbeyer:
Woha, so finally this gets merged? Feel free to do it yourself!

Issue #92, opened 2014-10-03
Add likely and unlikely macros


neithernut:
The likely and unlikely macros would be especially useful for the singletons.

PR #93, opened 2014-10-03
Add likely()/unlikely() macros

matthiasbeyer:
Targets #92


src/util/condition.h
```diff
@@ -0,0 +1,65 @@
...
+ * @{
+ */
+
+/**
+ * @addtogroup utils_condition "(internal) utilities"
+ *
+ * @{
+ */
+
+#ifndef __WS_UTIL_ARITHMETICAL_H__
```
neithernut: You forgot to adjust the include-guard.

neithernut:
ACK.

PR #94, opened 2014-10-04
Add testing setup

matthiasbeyer:
This is my proposal for the testing setup.

My Idea is that each module gets its own testing. No global test executable
to execute. Instead, we have a common testing helper source (tests.h), which
provides some utilities (by now just a macro for generating the `main()`
function) and each subdir in `tests/check/` is meant to produce an own
executeable test.

This way, we would have a much better testing setup and one could run only
the appropriate tests while developing.

I consider this setup as really clean. It is _inspired_ by the testing
setup of the gstreamer project, which also uses check.


matthiasbeyer:
What the hell? Something went horribly wrong. I'll close this.

matthiasbeyer:
For documentation: I use the `git-hub` tool for creating PRs from commandline.

The `git-hub` tool assumed `master` as the base branch for the pull request (because I did not specify `-b develop`) and here we go.

PR #95, opened 2014-10-04
Add test files

matthiasbeyer:
This is my proposal for the testing setup.

My Idea is that each module gets its own testing. No global test executable
to execute. Instead, we have a common testing helper source (tests.h), which
provides some utilities (by now just a macro for generating the main()
function) and each subdir in `tests/check/` is meant to produce an own
executeable test.

This way, we would have a much better testing setup and one could run only
the appropriate tests while developing.

I consider this setup as really clean. It is _inspired_ by the testing
setup of the gstreamer project, which also uses check.

### DIFF
--- a/tests/check/tests.h
+++ b/tests/check/tests.h
@@ -1,0 +1,68 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @addtogroup tests "Testing"
+ *
+ * This module contains all the test sources for the check framework based
+ * tests.
+ *
+ * @{
+ */
+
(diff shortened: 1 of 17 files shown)


tests/check/tests.h
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * @{
+ */
+
+#ifndef __WS_TESTS_TESTS_H__
+#define __WS_TESTS_TESTS_H__
+
+#define WS_TESTS_CHECK_MAIN_ADJUST(genf,ckflags)                   \
+    int                                                            \
+    main(                                                          \
```
neithernut: The `main()` function isn't that long, but I would move the contents to some utility function, taking `genf` and `ckflags` as parameters.

---

Do you actually use `genf`? Can's spot it...

matthiasbeyer: The point of the `main()` function in the macro is that each testing file has it's own `main()` function, which does execute the suite.

---

The call to `genf()` wasn't fixed up after my work, the commit I just added fixes this.

matthiasbeyer:
@neithernut Can you review the build setup carefully? I tried to implement it on my own, I hope I got it right!

matthiasbeyer:
Btw: I'd like to add an additional target for running all of the tests with one command, but I don't know how to do this properly.

I would add a entry in the `.drone.yml` file for it, then. Feel free to do this and send me a quick note from wich of your branches I can pull such changes after you added them (on your own will, of course).

tests/check/tests.h
neithernut: Of course each test executable has it's own name, but that's not my point.
Moving the code to a function on it's own keeps the macro short (like, 3 lines), which is always a good thing.

neithernut:
We just do the same we did in `libreset`. We call `ctest`, which in turn runs all the tests.

matthiasbeyer:
So, it seems that this works. @neithernut Is the test setup (the cmake stuff) clean the way I wrote it? Would appreciate your oppinion!

neithernut:
Looks good, you can, however, avoid code repetition in the `CMakeLists.txt` by setting up the targets in a `foreach`. If you want to, I can show you how.

matthiasbeyer:
Feel free to add a commit on top of this branch and notify me for it afterwards (without PR).

neithernut:
That's what I had in mind.

neithernut:
I pushed a branch with a bunch of commits.
Not ethat the test setup still lacks linkage and dependence on the other code.
There are two possibilities to solve this:
- Have a dummy library containing all the symbols we need, constructed from the source files in `src/`
- Add the source files directly

We have to link to all the libraries we need for the source to compile, too.

matthiasbeyer:
Feel free to add more commits.

I'd prefer the direct approach.

neithernut:
You have a GO for squashing.

neithernut:
I'll probably add a last addition anyways.

matthiasbeyer:
Done.

neithernut:
ACK.

PR #96, opened 2014-10-04
Add object logging

matthiasbeyer:
This PR adds the object logging mechanism after fixing some interface defintions for this.

### DIFF
--- a/src/objects/object.h
+++ b/src/objects/object.h
@@ -80,8 +80,8 @@ typedef bool (*ws_object_deinit_callback)(struct ws_object* const);
  * log callback
  */
 typedef bool (*ws_object_log_callback)(
-        struct ws_object* self,
-        struct ws_logger_context* const log_ctx
+        struct ws_logger_context* const log_ctx,
+        struct ws_object* self
 );
 
 /**
--- a/src/objects/object.h
+++ b/src/objects/object.h
@@ -79,7 +79,7 @@ typedef bool (*ws_object_deinit_callback)(struct ws_object* const);
 /**
  * log callback
  */
-typedef bool (*ws_object_log_callback)(
+typedef bool (*ws_object_dump_callback)(
         struct ws_logger_context* const log_ctx,
         struct ws_object* self
 );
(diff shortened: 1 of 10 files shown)


src/objects/object.h
```diff
@@ -75,10 +77,11 @@ typedef bool (*ws_object_deinit_callback)(struct ws_object* const);
 
 /**
  * log callback
- *
- * @todo Implement second parameter: struct ws_log_context*
  */
-typedef bool (*ws_object_log_callback)(struct ws_object* const, void*);
+typedef bool (*ws_object_log_callback)(
+        struct ws_object* self,
+        struct ws_logger_context* const log_ctx
```
neithernut: Context comes first.

matthiasbeyer:
Would it be beneficial to add a function which does literally the same but _always_ on `stderr`, and only including this function when compiling with debugging? So you could replace the vtable pointer and every call to `ws_object_log()` would cause debugging information.

A function for preventing a logging call would be nice, too (for debugging, of course). Something like `debug_dummy` or so.

What are you guys thinking about this? The overhead in the tests would be minimal, as you would only have to override the vtable ptr in the preface (is this actually possible when having constant vtables?)

neithernut:
Tests can ship replacement functions for logging anyways...

matthiasbeyer:
Indeed, so no need to add more code here.

neithernut:
Indeed.

neithernut:
Mh... do we have actually a usecase for this?

matthiasbeyer:
Mainly debugging.

matthiasbeyer:
Umm... but also normal logging. Something like "log that this monitor has changed, log that these windows were moved" etc.

neithernut:
I don't see how a logging function in _each_ object type would help. Printing the object pointer, hash and type would be sufficient. And that should be done using a regular logging call, not one embedded in the object model.

matthiasbeyer:
The callback in the object type is actually for doing deep-logging, printing attributes and so on. In case of a moved window this could be the new position for example.

neithernut:
So we dump the entire state of the object? Urgh...
In this case the name should be changed to `dump_state` or something.

matthiasbeyer:
A rename would break the build in various places, I guess.

neithernut:
I seriously doubt that. The only derived types are still in PRs.

matthiasbeyer:
Btw, @neithernut : The rename crashed the build as expected.

Please review again. And merge this if acknowledged.

neithernut:
ACK.

Issue #67, opened 2014-09-30
Add value type "nil"


neithernut:
We need `ws_value_nil`.

neithernut:
Blocks #64.

PR #97, opened 2014-10-04
Add implementation for value_nil type

matthiasbeyer:
Targets #67

### DIFF
--- a/src/values/nil.c
+++ b/src/values/nil.c
@@ -27,3 +27,21 @@
 
 #include "values/nil.h"
 
+#include "values/value.h"
+
+void
+ws_value_nil_init(
+    struct ws_value_nil* self
+) {
+    if (self) {
+        ws_value_init(&self->value);
+        self->value.type = WS_VALUE_TYPE_NIL;
+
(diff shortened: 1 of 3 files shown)


src/values/nil.c
```diff
@@ -27,3 +27,21 @@
...
+    if (self) {
+        ws_value_init(&self->value);
+        self->value.type = WS_VALUE_TYPE_NIL;
+
+        /*
+         * DON'T set the deinit callback, as the NIL type is nothing
+         * special. We don't need a deinit functionality, because it is
+         * just a value type with another name
+         */
+        //self->value.deinit_callback = nil_deinit;
```
neithernut, line 45: It has to be set to NULL, though.

matthiasbeyer, line 45: Actually not. The supertype cares about this and calls the deinit callback which was set beforehand.

But I found a endless-recursion bug because of your note, so thanks!

neithernut, line 45: k.

neithernut:
ACK.

Issue #68, opened 2014-09-30
Add value type "bool"


neithernut:
We need `ws_value_bool`.

PR #98, opened 2014-10-04
Add implementation for ws_value_bool type

matthiasbeyer:
Targets #68

### DIFF
--- a/src/values/bool.c
+++ b/src/values/bool.c
@@ -62,5 +62,11 @@ ws_value_bool_set(
     struct ws_value_bool* self,
     bool b
 ) {
-    return 0;
+    if (self) {
+        self->b = b;
+
+        return 0;
+    }
+
+    return -EINVAL;
 }
--- a/src/values/bool.c
+++ b/src/values/bool.c
@@ -48,6 +48,12 @@ bool
 ws_value_bool_plain(
     struct ws_value_bool* self
 ) {
+    if (self) {
+        bool b = self->b;
+
+        return b;
+    }
+
     return false;
 }
 
--- a/src/values/bool.c
+++ b/src/values/bool.c
@@ -25,5 +25,30 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
+#include <pthread.h>
+#include <stdbool.h>
+
 #include "values/bool.h"
 
+
+void
+ws_value_bool_init(
+    struct ws_value_bool* self
+) {
+}
+
+bool
+ws_value_bool_plain(
+    struct ws_value_bool* self
+) {
+    return false;
+}
+
+int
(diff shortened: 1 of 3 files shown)


src/values/bool.c
```diff
@@ -25,5 +25,65 @@
...
 
+/**
+ * Deinit callback for ws_value_bool type
+ */
+static void
+bool_deinit(
+    struct ws_value* self
+) {
+    if (self) {
+        if (self->type == WS_VALUE_TYPE_BOOL) {
```
neithernut: You can savely assume that the type is, indeed, `WS_VALUE_TYPE_BOOL`.

src/values/bool.h
```diff
@@ -34,7 +34,57 @@
...
+/**
+ * ws_value_bool type definition
+ */
+struct ws_value_bool {
+    struct ws_value value; //!< Supertype
+
+    struct {
+        bool b; //!< Actual value
+        pthread_rwlock_t rwl; //!< Read-Write lock for atomic operations
+    } boolean; //!< Thread-safe boolean type
```
neithernut: Why the encaplulation?

matthiasbeyer: For clearification what belongs to what.

neithernut: It is not neccessary. We _always_ want to lock the entire type, so we always lock the value.

---

Should we put the lock into the `value`?

matthiasbeyer: > Should we put the lock into the `value`?

I think so. Shall I open a new PR for this, or can it go in this one?

neithernut:
I'm not 100% sure abot this yet, as objects are lockable on their own.
Maybe we really should have the locks in the basic datatype values. However, the data should not be bundeled in an anonymous struct.

src/values/bool.h
neithernut: Forget about this. I was wrong here, as I commented somewhere else.

---

The lock should still be removed, though.

matthiasbeyer: So I guess I can remove the locking for the boolean (the actual `bool`), right?

neithernut: Yep.

matthiasbeyer:
I'm ready with this. ACK for squashing?

src/values/bool.c
```diff
@@ -25,5 +25,58 @@
...
+#include <stdbool.h>
+
+#include "util/attributes.h"
 #include "values/bool.h"
 
+/**
+ * Deinit callback for ws_value_bool type
+ */
+static void
+bool_deinit(
```
neithernut: Ame as for int. No `deinit` needed here.

neithernut:
GO for squashing.

matthiasbeyer:
Done.

neithernut:
ACK.

Issue #69, opened 2014-09-30
Add value type "int"


neithernut:
We need `ws_value_int`.

neithernut:
Blocks #64.

PR #99, opened 2014-10-04
Add implementation for ws_value_int type

matthiasbeyer:
Targets #69

### DIFF
--- a/src/values/int.c
+++ b/src/values/int.c
@@ -50,7 +50,14 @@ ws_value_int_set(
     struct ws_value_int* self,
     uint32_t i
 ) {
-    return 0;
+    if (self) {
+        if (self->value.type == WS_VALUE_TYPE_INT) {
+            struct ws_value_int* obj = (struct ws_value_int*) self;
+            obj->i = i;
+            return 0;
+        }
+    }
+    return -EINVAL;
 }
 
 uint32_t
--- a/src/values/int.c
+++ b/src/values/int.c
@@ -25,5 +25,38 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#include "values/int.h"
+#include <errno.h>
+#include <pthread.h>
+#include <stdint.h>
 
+#include "values/int.h"
+#include "values/value.h"
+
+void
+ws_value_int_init(
+    struct ws_value_int* self
+) {
+}
+
+int
+ws_value_int_set(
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
Note: Our specification says we should implement this one with 64 bit. I implemented it using 32 bit. What should be used?

src/values/int.c
```diff
@@ -25,5 +25,104 @@
...
+
+/**
+ * Deinitialize callback for ws_value_int type
+ */
+static void
+int_deinit(
+    struct ws_value* self
+) {
+    if (self) {
+        if (self->type == WS_VALUE_TYPE_INT) {
```
neithernut: You don't need this check.

src/values/int.c
```diff
@@ -25,5 +25,104 @@
...
+    struct ws_value_int* other
+) {
+    if (self && other) {
+        if (self->value.type == WS_VALUE_TYPE_INT &&
+                other->value.type == WS_VALUE_TYPE_INT) {
+
+            signed int res;
+
+            pthread_rwlock_rdlock(&self->i.rwl);
+            pthread_rwlock_rdlock(&other->i.rwl);
```
neithernut: You created a potential deadlock.
Use the `try` variant and potentially a condition or something.

matthiasbeyer: I guess a `try` lock wouldn't be beneficial here, but I don't know how to properly use conditions (by now).

Would you care to add a patch on this branch? If not, you are welcome to teach me how to implement it tomorrow evening!

neithernut: It'll be tomorrow evening then.

neithernut: As discussed somewhere else, remove the locks completely.

matthiasbeyer:
I'm ready here, ACK for squashing?

src/values/int.c
```diff
@@ -25,5 +25,87 @@
...
+
 #include "values/int.h"
+#include "values/value.h"
+#include "util/attributes.h"
+
+/**
+ * Deinitialize callback for ws_value_int type
+ */
+static void
+int_deinit(
```
neithernut: Set the deinit callback to ``NULL and be done with it. No need fo a NOP function.

neithernut:
GO for squashing.

matthiasbeyer:
Done.

neithernut:
ACK.

Issue #73, opened 2014-09-30
Add value type "named"


neithernut:
We need `ws_value_named`.

neithernut:
Blocks #64.

PR #100, opened 2014-10-04
Add implementation for ws_value_named_value type

matthiasbeyer:
Targets #73

### DIFF
--- a/src/values/value_named.c
+++ b/src/values/value_named.c
@@ -25,5 +25,56 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
+#include <pthread.h>
+#include <string.h>
+
+#include "objects/object.h"
+#include "objects/string.h"
+
+#include "values/value.h"
 #include "values/value_named.h"
 
+void
+ws_value_named_value_init(
+    struct ws_value_named_value* self
+) {
+}
+
+int
+ws_value_named_set_name(
+    struct ws_value_named_value* self,
+    char * str //!< new name
+) {
+    return 0;
+}
+
+int
+ws_value_named_set_value(
+    struct ws_value_named_value* self,
+    struct ws_value* val
+) {
+    return 0;
+}
+
+char const*
+ws_value_named_get_name(
+    struct ws_value_named_value* self
+) {
+    return NULL;
+}
+
+struct ws_value*
+ws_value_named_get_value(
+    struct ws_value_named_value* self
+) {
+    return NULL;
+}
+
+signed int
+ws_value_named_cmp(
+    struct ws_value_named_value* self,
+    struct ws_value_named_value* other
+) {
+    return 0;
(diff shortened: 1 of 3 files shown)


src/values/value_named.h
```diff
@@ -34,7 +34,101 @@
...
+/**
+ * ws_value_named_value type definition
+ */
+struct ws_value_named_value {
+    struct ws_value value; //!< Supertype
+
+    struct {
+        char* str;
+        pthread_rwlock_t rwl;
+    } name; //!< Name of the value, ascii
```
neithernut: Would use `ws_string` here.

matthiasbeyer: The spec notes explicitely that this one is only ASCII.

neithernut: Yes, but it makes using named values from a transaction much easier.

matthiasbeyer: We agreed on ASCII because we wanted the restriction for the API that it only can use ASCII characters for naming values (variables). It prevents us from the `ws_string` overhead, also.

matthiasbeyer:
What about this, @neithernut ?

src/values/value_named.c
```diff
@@ -25,5 +25,129 @@
...
+#include "objects/string.h"
+
+#include "values/value.h"
 #include "values/value_named.h"
 
+static void
+deinit_namedval(
+    struct ws_value* self
+) {
+    if (self) {
```
neithernut: _please_ do an early return here. And merge the `if`s.

src/values/value_named.c
```diff
@@ -25,5 +25,129 @@
...
+            ws_object_unref((struct ws_object*) nv->name);
+        }
+    }
+}
+
+void
+ws_value_named_value_init(
+    struct ws_value_named_value* self
+) {
+    if (self) {
```
neithernut: Early return.

src/values/value_named.c
```diff
@@ -25,5 +25,129 @@
...
+    return NULL;
+}
+
+signed int
+ws_value_named_cmp(
+    struct ws_value_named_value* self,
+    struct ws_value_named_value* other
+) {
+
+    /*
```
neithernut, line 120: Instead of multiline comments, use one line comments in the places where you _do_ things.

matthiasbeyer, line 120: Why? This is just a minor styling thing ...

neithernut, line 120: It enhances readability.

matthiasbeyer, line 120: That's just a personal preference IMO.

TheNeikos, line 120: [All I have to say](http://www.wikiwand.com/en/Parkinson's_law_of_triviality)

matthiasbeyer:
So what is the conclusion on this?

neithernut:
ACK from me.

PR #101, opened 2014-10-04
Fixes for the values/value.h header

matthiasbeyer:
Blocks #97, #98, #99, #100

PR #105, opened 2014-10-05
Fix: ws_value implementations

matthiasbeyer:
This is a result of my stupidity. #101 gets fixed by this and
_this time the CI actually builds the code_.


neithernut:
ACK.

PR #106, opened 2014-10-05
Fix: Potential endless recursion bug

matthiasbeyer:
Fix: Potential endless recursion bug if `deinit_callback` member
is set to `ws_value_deinit`.


neithernut:
Please add an `unlikely`.

matthiasbeyer:
You mean `likely`.

neithernut:
Ah, yeah.

Issue #118, opened 2014-10-06
ws_object_unlock_read and ws_object_unlock_write should be merged


neithernut:
Both do exactly the same thing.
There should only be `ws_object_unlock`

PR #121, opened 2014-10-06
Add ws_object_unlock() and make old unlock methods deprecated

matthiasbeyer:
Targets #118

### DIFF
--- a/src/objects/array.c
+++ b/src/objects/array.c
@@ -117,7 +117,7 @@ ws_array_sort(
     if (self) {
         ws_object_lock_write(&self->obj);
         qsort(self->ary, sizeof(*self->ary), self->len, cmp);
-        ws_object_unlock_write(&self->obj);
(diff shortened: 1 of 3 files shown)


neithernut:
Add the deprecated attributes.
Also: we should ultimately remove them as soon as possible (definitely before v0.1)

matthiasbeyer:
I cannot force people working on other branches to switch to the new interface, so the deprecated flag/attribute will be a notice and yes, should be fixed until v0.1 or something.

matthiasbeyer:
Note: When merging other PRs which still use the old functions, the CI doesn't fail, but after the merge it failes. That's because the CI doesn't do a merge before building. So reviewing the other PRs should be done _really_ carefully!

neithernut:
Or they should be rebased.

neithernut:
Btw: ACK.

PR #122, opened 2014-10-06
Make ws_value a lockable type

matthiasbeyer:
Blocks #97, #98, #99, #100


neithernut:
As noted in another PR. I'm not yet quite sure about this...

matthiasbeyer:
Why not?

matthiasbeyer:
There is actually one point why we wouldn't need a locking mechanism for `ws_value` types: They are supposed to be used only in the execution of transactions.

But in fact, they won't. They can also be used to identify transactions, pass values around between compositor (model, actually) and transactions and so on. So they need to be thread safe. For example, the location of a window will be represented by an object (or the complete window). Transactions only know about the `ws_object` instance because of the `ws_value_obref` (or whatever the name was) reference to it. And only _one_ thread should talk to the object at a time, whereas the compositor itself should still be able to read/write the underlying `ws_object` instance. So the very first approach which comes to my mind is that we need to lock the _reference_ on the actual _instance_ here, so transactions will do (kindof) atomic operations while the compositor is still able to do operations on the actual object.

This is at least my point of view, you may have others and this isn't meant to be offensive or something. Just my thoughts why this should go in there. It's not that much overhead, though.

What's your opinion?

PS: Others are welcome to argue here! @TheNeikos , @manuelmessner , @xVin , @neithernut

matthiasbeyer:
Question: In the PR, I do return `int` as status of the lock operation. In the `ws_object` type, we return `bool`. Which one should be adapted? 

(I'd suggest this one, as changing the `ws_object` interface again would again cause a lot of changes in other places and so on)

neithernut:
Locking the reference is completely and utterly pointless.

And if we pass values around, we will end up having copies.
I'd rather secure the way those copies are moved (ass in, we make a copy in one thread, then stop modifying it and pass it to another) than having locks. Of course, we _should_ have a look at memory barriers for some cases, but we're completely save doing things that way. It's far more efficient, too.

Conclusion: complex objects are thread save on their own and we don't have to care about basic values (ints and bools) as values will not be used from more than one thread at a time. This should ,however, be documented.

This results in this PR to be closed unmerged, of course.

Issue #113, opened 2014-10-06
Make framebuffers/monitors objects


neithernut:
The framebuffers and connectors should be made objects, which get created by the compositor upon initialisation.

TheNeikos:
Taking this one!

neithernut:
Blocks #126.

PR #125, opened 2014-10-07
Add framebuffer object and monitor object

TheNeikos:
This implements the framebuffer and monitor object

### DIFF
--- a/src/compositor/framebuffer_device.c
+++ b/src/compositor/framebuffer_device.c
@@ -26,18 +26,18 @@
  */
 
 #include <errno.h>
+#include <fcntl.h>
 #include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
 #include <unistd.h>
 #include <xf86drm.h>
-#include <fcntl.h>
-#include <unistd.h>
-#include <string.h>
 
 #include "logger/module.h"
 #include "objects/object.h"
 
-#include "compositor/internal_context.h"
 #include "compositor/framebuffer_device.h"
+#include "compositor/internal_context.h"
 
 
 /*
--- a/src/compositor/framebuffer_device.c
+++ b/src/compositor/framebuffer_device.c
@@ -1,0 +1,125 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <xf86drm.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+
+#include "logger/module.h"
+#include "objects/object.h"
+
+#include "compositor/internal_context.h"
+#include "compositor/framebuffer_device.h"
+
+
+/*
+ *
+ *  Forward declarations
+ *
+ */
+
+static bool
+ws_framebuffer_device_deinit(
+    struct ws_object* self
+);
+
+static size_t
+ws_framebuffer_device_hash(
+    struct ws_object* obj
+);
+
+/*
+ *
+ * Interface Implementation
+ *
+ */
+
+ws_object_type_id WS_OBJECT_TYPE_ID_FRAMEBUFFER_DEVICE = {
+    .supertype = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr = "ws_framebuffer_device",
+    .deinit_callback = ws_framebuffer_device_deinit,
+    .hash_callback = ws_framebuffer_device_hash,
+    .init_callback = NULL,
+    .log_callback = NULL,
+    .run_callback = NULL
+};
+
+struct ws_framebuffer_device*
+ws_framebuffer_device_new(
+    char* path
+) {
+    struct ws_framebuffer_device* tmp = calloc(1, sizeof(*tmp));
+    ws_object_init((struct ws_object*)tmp);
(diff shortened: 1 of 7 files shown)


neithernut:
Targets issue #113.

src/compositor/framebuffer_device.c
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <errno.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include <xf86drm.h>
+#include <fcntl.h>
```
matthiasbeyer: Please sort them.

src/compositor/framebuffer_device.c
```diff
@@ -0,0 +1,125 @@
...
+#include <unistd.h>
+#include <string.h>
+
+#include "logger/module.h"
+#include "objects/object.h"
+
+#include "compositor/internal_context.h"
+#include "compositor/framebuffer_device.h"
```
matthiasbeyer, line 39: Sort them here, too.

src/compositor/framebuffer_device.h
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ * ws_framebuffer type definition
+ *
+ * The ws_framebuffer
+ */
+struct ws_framebuffer_device {
+    struct ws_object obj; //!< Supertype
+    int fd; //!< The filedescriptor
+    char* path; //!< The path to the descriptor
```
matthiasbeyer: To the descriptor?

src/compositor/internal_context.h
```diff
@@ -42,9 +36,10 @@ struct ws_framebuffer_device {
  * This context holds the internal state of the compositor.
  */
 extern struct ws_compositor_context {
-    struct ws_framebuffer_device fb;
+    struct ws_framebuffer_device* fb;
     struct ws_monitor* conns; //<! A linked list of ws_monitors
 } ws_comp_ctx;
 
+extern struct ws_logger_context log_ctx;
```
matthiasbeyer, line 44: Can you add documentation why you do this?

src/compositor/module.c
```diff
@@ -44,7 +44,7 @@
 #include "background_surface.h"
 
 struct ws_compositor_context ws_comp_ctx;
-static struct ws_logger_context log_ctx = { "[Compositor] " };
+struct ws_logger_context log_ctx = { "[Compositor] " };
```
matthiasbeyer: I would suggest you specify the member:

``` C
= { .prefix = "text" };
```

src/compositor/monitor.c
```diff
@@ -0,0 +1,130 @@
...
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include <errno.h>
+#include <stdlib.h>
```
matthiasbeyer: Sort them, please!

src/compositor/monitor.h
```diff
@@ -0,0 +1,96 @@
...
+ * @addtogroup objects_monitor "Class: Monitor type"
+ *
+ * @{
+ */
+
+#ifndef __WS_OBJECTS_MONITOR_H__
+#define __WS_OBJECTS_MONITOR_H__
+
+#include <stdbool.h>
+#include <pthread.h>
```
matthiasbeyer, line 43: Sort them, please!

src/compositor/monitor.h
```diff
@@ -0,0 +1,96 @@
...
+
+#include "objects/object.h"
+#include <xf86drmMode.h>
```
matthiasbeyer: This should go to the upper headers, the non-internal ones.

src/compositor/framebuffer_device.h
TheNeikos: Yes, the filedescriptor just above

matthiasbeyer: But it is not the path to the actual descriptor but the "thing" (file) behind it...

TheNeikos: And what is the difference?

A fd to `/dev/null` will still have a path to `/dev/null`

matthiasbeyer: But the fd itself and the thing, the fd references are two different things, and your wording in the comment makes one think that they _are_ the same, at least from my understanding.

fd => reference to a file
path => path to reference to a file

But it should be

fd => reference to a file
path => path to the thing, the fd references

neithernut:
GO for squashing.

src/compositor/framebuffer_device.c
```diff
@@ -0,0 +1,142 @@
...
+}
+
+static int
+ws_framebuffer_device_cmp(
+    struct ws_object const* obj1,
+    struct ws_object const* obj2
+) {
+    struct ws_framebuffer_device* dev1 = (struct ws_framebuffer_device*) obj1;
+    struct ws_framebuffer_device* dev2 = (struct ws_framebuffer_device*) obj2;
+    return !(dev1->fd == dev2->fd);
```
matthiasbeyer: The cmp  call should return -1, 0 or 1, not 1/0 true/false!

src/compositor/monitor.c
```diff
@@ -0,0 +1,147 @@
...
+}
+
+static int
+ws_monitor_cmp(
+    struct ws_object const* obj1,
+    struct ws_object const* obj2
+) {
+    struct ws_monitor* mon1 = (struct ws_monitor*) obj1;
+    struct ws_monitor* mon2 = (struct ws_monitor*) obj2;
+    return !(mon1->crtc == mon2->crtc && mon1->fb_dev->fd == mon2->fb_dev->fd);
```
matthiasbeyer: Also here, cmp should return -1, 0, 1.

TheNeikos: I can't. It doesn't just compare a _single_ value but two.

src/compositor/monitor.h
```diff
@@ -0,0 +1,96 @@
...
+    struct ws_object obj; //!< Supertype
+    int connected; //!< is the monitor connected?
+    uint32_t width; //!< width of the buffer
+    uint32_t height; //!< height of the buffer
+    uint32_t stride; //!< length of a line
+
+    uint32_t size; //!< size of the map
+    uint8_t* map; //!< mapped buffer data
+
+    struct ws_framebuffer_device* fb_dev; //!< Framebuffer Device
```
matthiasbeyer, line 64: Is there a point why this is not embedded in the `ws_monitor` type?

TheNeikos, line 64: Because there should be only one (per device)! **To rule them all!**

src/compositor/monitor.c
matthiasbeyer: You could return `0` if everything is equal and do some more calculation in the other case.

I guess you have to find a nice algorithm for this, because the compare functions needs the appropriate return.

matthiasbeyer: Btw: Is this a callback? Callbacks (which are not exported, btw.) shouldn't be named with the `ws_` prefix.

src/compositor/monitor.h
matthiasbeyer, line 64: :+1: :smile:

neithernut:
ACK.

PR #130, opened 2014-10-08
Remove the init callback for the array type

matthiasbeyer:
This PR removes the init callback for the array type. Init callbacks
do not make sense and we cannot use them as expected, so this is a
bugfix in some way.


matthiasbeyer:
@neithernut ?

neithernut:
ACK.

PR #133, opened 2014-10-08
Changed monitor creation to sets

TheNeikos:
This commit changes the monitor list into a set, allowing for an easier
interface.

It changes populate_framebuffer so as to fit into a ws_set_select call. As this
is the way to iterate over the set.

Find connector now also searches using a dummy object instead of iterating
over a linked list.


matthiasbeyer:
You always work with commenting out code fragments. I would highly appreciate a comment (right in the source) about _why_ the fragment got outcommented.

Makes maintaining much easier.

TheNeikos:
You are right, adding something now

matthiasbeyer:
Looks good to me, but @neithernut should review that, too!

neithernut:
ACK.

PR #136, opened 2014-10-08
Moved the framebuffer population to the monitor

TheNeikos:
This commit moves the populate_framebuffers to the monitor internals.
This was planned from the start and thus the code did not really change and
only the return values had to be changed.


TheNeikos:
This is for https://github.com/waysome/waysome/issues/126

neithernut:
ACK.

src/compositor/module.c
```diff
@@ -357,73 +357,6 @@ populate_framebuffers(
...
-
-    return 0;
-err_fb:
-    drmModeRmFB(ws_comp_ctx.fb->fd, monitor->fb);
-err_destroy:
-    memset(&dreq, 0, sizeof(dreq));
-    dreq.handle = monitor->handle;
-    drmIoctl(ws_comp_ctx.fb->fd, DRM_IOCTL_MODE_DESTROY_DUMB, &dreq);
-
+    ws_monitor_populate_fb(monitor);
```
matthiasbeyer, line 360: So, `populate_framebuffers()` does _nothing_ besides casting and calling this new function on the casted thing. Why not removing `populate_framebuffers()` completely?

TheNeikos, line 360: Because it makes sense. You cannot put `ws_monitor_populate_fb` into a `set_select`, thus you need another function to do it for you.

neithernut, line 360: `set_select` needs a function which takes 2 arguments.

matthiasbeyer, line 360: Ok.

matthiasbeyer:
ACK.

Issue #110, opened 2014-10-06
Implement buffer interface


neithernut:
We need to implement the buffer interface, which will serve as the base for drm- and shared-memory-buffers.

PR #140, opened 2014-10-08
Add compositor wayland buffer

neithernut:
This PR introduces the wayland buffer, which is used to get image data from applications.

### DIFF
--- a/src/compositor/wayland_buffer.c
+++ b/src/compositor/wayland_buffer.c
@@ -25,6 +25,9 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
+#include <malloc.h>
+
 #include "compositor/wayland_buffer.h"
 
 ws_object_type_id WS_OBJECT_TYPE_ID_WAYLAND_BUFFER = {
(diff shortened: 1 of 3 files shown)


neithernut:
Targets issue #110.

src/compositor/wayland_buffer.c
```diff
@@ -0,0 +1,312 @@
...
+ *
+ * Module specific utility macros
+ *
+ */
+
+/**
+ * Get the ws_wayland_buffer for a ws_buffer
+ */
+#define mother(self) ((struct ws_wayland_buffer*) \
+                     ((void*) (self))-offsetof(struct ws_wayland_buffer, buf))
```
matthiasbeyer: Nope. Just Nope. I suggest we use the `container_of` macro, the linux kernel has and put it into our utils, adding a nice alias macro here, but I will not ACK this line of code.

src/compositor/wayland_buffer.c
```diff
@@ -0,0 +1,312 @@
...
+ *
+ */
+
+/**
+ * Buffer type
+ */
+static ws_buffer_type_id buffer_type = {
+    .type = {
+        .supertype  = (struct ws_object_type const*) &WS_OBJECT_TYPE_ID_BUFFER,
+        .typestr    = "anonymous",
```
matthiasbeyer, line 119: Why no real name here?

neithernut, line 119: Because it's actually an internal buffer which is only exported.

src/compositor/wayland_buffer.c
neithernut: No need, wayland ships `wl_container_of`. And it does _just_ the same (using `offsetof`).

matthiasbeyer: So why not using this one?

neithernut: I changed that. But the macro is kind of contra-continutive.

matthiasbeyer:
ACK for squashing from me.

neithernut:
No need/Done.

PR #141, opened 2014-10-08
Use the method of a base if it's unset

neithernut:
After applying this PR, if you set a callback to `NULL`, the system will search for implementation i base classes.
Deinit and some other special methods are not affected by this.

### DIFF
--- a/src/compositor/buffer.c
+++ b/src/compositor/buffer.c
@@ -71,8 +71,14 @@ ws_buffer_data(
     struct ws_buffer* self
 ) {
     ws_buffer_type_id* type = (ws_buffer_type_id*) &self->obj.id;
-    if (!(type && type->get_data)) {
-        return NULL;
+
+    // search for an implementation in the base classes
+    while (!type->get_data) {
+        // we hit the basic, abstract buffer type, which does nothing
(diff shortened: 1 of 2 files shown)


src/compositor/buffer.c
```diff
@@ -71,8 +71,13 @@ ws_buffer_data(
     struct ws_buffer* self
 ) {
     ws_buffer_type_id* type = (ws_buffer_type_id*) &self->obj.id;
-    if (!(type && type->get_data)) {
-        return NULL;
+
+    // search for an implementation in the base classes
+    while (!type->get_data) {
+        if (type == &WS_OBJECT_TYPE_ID_BUFFER) {
```
matthiasbeyer, line 78: You should add a note why we stop on the `WS_OBJECT_TYPE_ID_BUFFER` type identification! I mean, it is clear now, but maybe not later.

matthiasbeyer, line 78: As well for the other occurences, of course.

neithernut:
Do I have a GO for squashing?

matthiasbeyer:
From me: yes.

neithernut:
Done.

Issue #129, opened 2014-10-08
Implement ws_connbuf


neithernut:
We'll need a `ws_connbuf` at some point (for the `ws_connector`).

matthiasbeyer:
For clearification: Should this be an `ws_object` subtype?

neithernut:
Nope.

neithernut:
Blocks #160.

PR #143, opened 2014-10-08
Added ws_connbuf type

manuelmessner:


manuelmessner:
Targets #129

matthiasbeyer:
For future submissions: You can put the "Targets xyz" in the description.

matthiasbeyer:
Is this supposed to be an singleton? If not, there are init/deinit functions missing!

matthiasbeyer:
ACK for implementing, from me. @neithernut ?

neithernut:
Yep.

src/connection/connbuf.h
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * Allocate a dynamic buffer and return the address
+ *
+ * @memberof ws_connbuf
+ *
+ * @return The address of the allocated memory on success, else NULL
+ */
+char*
+ws_connbuf_reserve(
+    int amount  //!< The amount how much memory will be allocated
```
matthiasbeyer: Why is this an `int` while the member in the struct is a `size_t`? Wouldn't it be sufficient to have a `size_t` here as well?

src/connection/connbuf.h
```diff
@@ -0,0 +1,109 @@
...
+);
+
+/**
+ * Increase the `size` member of the ws_connbuf struct
+ *
+ * @memberof ws_connbuf
+ */
+void
+ws_connbuf_append(
+    int amount //!< The amount how much the `size` member will be increased
```
matthiasbeyer: Same here: Why not `size_t`?

src/connection/connbuf.h
```diff
@@ -0,0 +1,109 @@
...
+
+/**
+ * Releases a given amount of memory and decrease the `amount` and `size`
+ * members
+ *
+ * @memberof ws_connbuf
+ */
+void
+ws_connbuf_discard(
+    int amount //!< the amount of memory to release
```
matthiasbeyer: And here: Why not `size_t`?

src/connection/connbuf.h
matthiasbeyer: Also: Is the connbuf type a singleton? If not, you need to pass the object itself as the first argument (also for the following functions).

The return type should change too, then.

src/connection/connbuf.h
```diff
@@ -0,0 +1,109 @@
...
+    size_t size; //!< Size of the allocated memory
+    size_t data; //!< Amount of the used memory
+};
+
+/**
+ * Initializes a ws_connbuf struct
+ *
+ * @memberof ws_connbuf
+ *
+ * @warning The `buffer` itself will not be allocated
```
matthiasbeyer: Why?

matthiasbeyer:
Q: Wouldn't it be convenvient to provide a method `ws_connbuf_new_with()` to allocate a connbuf directly with a certain amount of memory? (This question is also for @neithernut )

src/connection/connbuf.h
neithernut: For the third time: `ws_connbuf` is _not_ a singletin.

neithernut:
Could be done, but it isn't exactely neccessary.

src/connection/connbuf.c
```diff
@@ -0,0 +1,95 @@
...
+void
+ws_connbuf_deinit(
+    struct ws_connbuf* obj
+) {
+    if (!obj) {
+        return;
+    }
+
+    free(obj->buffer);
+    free(obj);
```
neithernut: The connbuf will be embedded in another struct, so do not free it (the method's name is `deinit` anyway).

src/connection/connbuf.c
```diff
@@ -0,0 +1,95 @@
...
+
+char*
+ws_connbuf_reserve(
+    int amount
+) {
+    if (amount < 1) {
+        return NULL;
+    }
+
+    return calloc(amount, sizeof(char));
```
neithernut: Er... it's not supposed to work this way.
A connbuf has a fixed amount of memory for the buffer.
- `ws_connbuf_reserve`ing memory of that buffer is semantically the same as marking the next `amount` bytes as being reserved for being filled by a writing entity.
- `ws_connfub_append` will then mark `amount` of those previously reserved bytes as written and resets the reservation.
- `ws_connbuf_discard` will take `amount` of those written bytes and discard them, as in, move the stuff not yet discarded to the beginning of the buffer.

So you have _one_ buffer, with three different areas:
- written bytes
- reserved bytes
- free bytes

You can discard written bytes, write to the reserved bytes and reserve free bytes.
I hope I have made sense here. If not, I can still pull out some paper and draw a few graphs on monday.

src/connection/connbuf.h
matthiasbeyer: This is nowhere documented, so how the hell should I remember it all the time?

neithernut: Read the spec?

matthiasbeyer: This is not stated in the docs explicitely.

Anyways, my first comment still holds.

neithernut: It _is_ in the specs! Have a look at the UML class diagrams. They say that two `ws_connbuf`s are part of a `ws_connetor`, which clearly implies that `ws_connbuf` is **no sinlgeton**.

src/connection/connbuf.c
manuelmessner: @matthiasbeyer 
I used the type int, because that's the way it was specified in the spec.
I was wondering myself, too (I wanted to ask @neithernut today).

@neithernut 
Ok, I rewrite it.

neithernut: You can use `size_t`.

manuelmessner:
@neithernut 
Can you review it please?

neithernut:
ACK @ header

manuelmessner:
@neithernut, @matthiasbeyer 
Review pls

src/connection/connbuf.c
```diff
@@ -0,0 +1,123 @@
...
+    struct ws_connbuf* self,
+    size_t amount
+) {
+    if (amount == 0) {
+        return NULL;
+    }
+
+    self->blocked = true;
+
+    if (self->size - self->data < amount) {
```
neithernut: `self->data + amount > self->size` would be more intuitive and it would also be safer regarding signedness.

src/connection/connbuf.c
```diff
@@ -0,0 +1,123 @@
...
+) {
+    if (amount == 0) {
+        return -EINVAL;
+    }
+
+    if (self->blocked) {
+        return -EINTR;
+    }
+
+    self->data -= amount;
```
neithernut, line 127: Please check whether `amount > data`.

;w
```diff
@@ -0,0 +1,128 @@
...
+    struct ws_connbuf* self,
+    size_t amount
+) {
+    if (amount == 0) {
+        return NULL;
+    }
+
+    self->blocked = true;
+
+    if (self->data + amount <= self->size) {
```
neithernut: I think this should be a `>`, not a `<=`.

src/connection/connbuf.c
matthiasbeyer: And parentheses!

;w
matthiasbeyer: And parentheses!

matthiasbeyer:
The file named ";w" should be something else, shouldn't it?

matthiasbeyer:
I think you should clean up what you did here and we will review afterwards. The comments in your diff are also for the ";w" file... which means that @neithernut did not even recognize that you fucked up here!

manuelmessner:
Yeah I fucked up yesterday -.-
I'll clean that up

matthiasbeyer:
Go ahead and clean your history.

neithernut:
Mh... seems you have to rebase.

src/CMakeLists.txt
```diff
@@ -34,8 +34,8 @@ add_definitions(
     ${PNG_DEFINITIONS}
     ${UDEV_DEFINITIONS}
     ${WAYLAND_CURSOR_DEFINITIONS}
-    ${WAYLAND_SERVER_DEFINITIONS}
     ${YAJL_DEFINITIONS}
+    ${WAYLAND_SERVER_DEFINITIONS}
```
neithernut, line 39: This is an unneccessary change.

neithernut:
Still waiting for the update...
Didn't you say you'd push it as soon as you had internet access, like, two hours ago?

manuelmessner:
@matthiasbeyer @neithernut  
Review pls

neithernut:
ACK

neithernut:
Better merge this before the commits are lost in a mishap.

matthiasbeyer:
@neithernut Commits are never lost. I'd highly appreciate if you wouldn't do such things!

neithernut:
They don't get lost if they are present in an uncorrupted repo. However, they _can_ get lost if, somehow, all the repos containing the commits get corrupted. This _is_, of course, unlinkely. My remark was actually a plain joke at @manuelmessner's expense.

Also: I was waiting for this to be merged _for ages_, as this did block a lot of things needed for the current milestone. The code was carefully reviewed and all we were waiting for was the cleanup of the history. I did a review, again, just to be sure before merging. So this is in accordance to our workflow and I honestly don't see anything wrong in merging a branch in such a situation.

matthiasbeyer:
Sounds reasonable to me. Just be careful in such cases. I trust you.

PR #150, opened 2014-10-09
Fix copyf() in set for next version of libreset

matthiasbeyer:
The next version uses `void*` instead of `void const*` as
return/argument type.


neithernut:
ACK.

PR #155, opened 2014-10-09
[WIP] Add serializer interface

matthiasbeyer:
Targets #161.


src/command/command_argument.h
```diff
@@ -0,0 +1,43 @@
...
+#define __WS_COMMAND_TRANSACTION_COMMAND_ARGUMENT_H__
+
+#include "objects/object.h"
+
+/**
+ * Transaction command argument type
+ */
+struct ws_transaction_command_argument {
+    struct ws_object obj;
+    struct ws_value* argv;
```
neithernut: We have a stack/dynamic range of registers. You don't want new values but the addresses of existing registers.
Use an array of `size_t` or some ints (add the length or come up with some termination symbol)

src/command/transaction.h
```diff
@@ -0,0 +1,43 @@
...
+#define __WS_COMMAND_TRANSACTION_H__
+
+#include "objects/object.h"
+#include "objects/array.h"
+
+/**
+ * Transaction type, extends the array type
+ */
+struct ws_transaction {
+    struct ws_object ary;
```
neithernut: You meant to have a `struct ws_array` here, didn't you?

matthiasbeyer: Jep.

src/command/command.h
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Transaction command
+ */
+struct ws_transaction_command {
+    struct ws_object obj;
+
+    char* callstr;
+
+    struct {
+        size_t ary;
```
neithernut: How is this supposed to hold multiplel params' positions?

matthiasbeyer: Argh.

matthiasbeyer:
@neithernut functions missing?

src/command/command.h
```diff
@@ -0,0 +1,50 @@
...
+#define __WS_COMMAND_TRANSACTION_COMMAND_H__
+
+#include "objects/object.h"
+#include "objects/array.h"
+
+/**
+ * Transaction command
+ */
+struct ws_transaction_command {
+    struct ws_object obj;
```
neithernut: A single command does not need to be an object.
If it's only for the sake of storing it in an array, you might use a `wl_array` instead for the transaction.

src/serialize/module.h
```diff
@@ -28,6 +28,42 @@
...
 
+#include <stddef.h>
 
+#include "objects/object.h"
+
+/**
+ * Serializer type
+ */
+struct ws_serializer {
+    struct ws_object obj; //!< Supertype
```
neithernut: A serializer isn't neccessarily an object.
We have exactely one serializer per command processor. We may have different types of deserializers at some point, but even in that case having a struct with callback fits better than making it an object.

src/command/command.h
```diff
@@ -0,0 +1,50 @@
...
+
+    char* callstr;
```
neithernut, line 38: It would be better to resolve this in the serializer.
Add a new function typedef like

``` C
int (*command)(struct ws_arguments const*)
```

Of course, to do this you have to make the struct below unanonymous.

neithernut:
I'd like to have this merged once the abstract definition and some basic infrastructure are done. This way, we can do this in paralell.

src/serialize/module.h
matthiasbeyer: We discussed this yesterday and yuo said it is an object!

src/command/command.h
matthiasbeyer, line 38: For what exactly do you wnat me to add a callback here?

src/command/command.h
matthiasbeyer: The transaction is already an array. So I guess we can just remove the inheritance and then we're fine here?

src/command/command.h
```diff
@@ -0,0 +1,50 @@
...
+
+    struct {
+        size_t* ary;
```
matthiasbeyer, line 41: Shouldn't we use fixed-size elements here? I would like to suggest `uint32_t`.

src/serialize/module.h
neithernut: I think I said that it is not derived from `ws_object`, but that it has an internal state, which serves as a context for yaijl.
object != object.

src/command/command.h
neithernut: Yep. But we discussed whether `ws_array`s hold `ws_objects` or other data, too, if you remember.

matthiasbeyer:
I added some more infrastructure and rewrote almost the complete history. The inheritance was split up once again and now we have:
- A serializer type
- A serializer backend type (abstract)
- A serializer serialize backend type (abstract)
- A serializer deserialize backend type (abstract)

The (json-, xml-, foobar-) backend (both serialize and deserialize) types are supposed to inherit from the latter two.

I'm not completely sure whether the last point is sane, as serializing and de-serializing could happen on the same backend object, _but_ what I had in mind with this is, that serializing and deserializing is a completely different task and with this setup, we could have different instances of the serializer doing serializing and deserializing _in parallel_, so a deserializing-job does not necessarily block a serializing job and the other way round. So this setup supports the threading and therefor the speed of the application.

---

More patches are coming.

src/command/command.h
```diff
@@ -0,0 +1,48 @@
...
+#include "objects/array.h"
+
+/**
+ * Transaction command
+ */
+struct ws_transaction_command {
+    char* callstr;
+
+    struct {
+        size_t* ary;
```
matthiasbeyer, line 41: I'll re-add this comment here: Shouldn't we use a fixed-size type, like `uint32_t`? Plugins would be much more portable this way!

neithernut, line 41: The pointer alone means you are bound the the concrete architecture concerning binary representation. `size_t` is defined for one architecture, so it's sane.

neithernut:
Introducing the possibility to perform serialisation and deserialisation in different threads is a good idea. However, I don't see any point in having them both based on a general serializer backend.

src/command/transaction.h
```diff
@@ -0,0 +1,43 @@
...
+#define __WS_COMMAND_TRANSACTION_H__
+
+#include "objects/object.h"
+#include "objects/array.h"
+
+/**
+ * Transaction type, extends the array type
+ */
+struct ws_transaction {
+    struct ws_array ary;
```
neithernut: As described earlier: making this a `wl_array` makes more sense.

src/serialize/backend.h
```diff
@@ -0,0 +1,112 @@
...
+ * (abstract) Serializer backend type
+ *
+ * @extends ws_object
+ *
+ * @note Serializer backends can be run.
+ *
+ * This is the supertype of all backend types for the serializer
+ */
+struct ws_serializer_backend {
+    struct ws_object obj; //!< @protected Base class.
```
neithernut, line 47: As described before: neither serializers nor a deserializers are _this kind of object_.

src/command/transaction.h
matthiasbeyer: It _is_ a `wl_array`!

src/serialize/backend.h
```diff
@@ -0,0 +1,112 @@
...
+
+    struct ws_transaction* t; //!< @protected Backend-Related transaction
+
+    struct {
+        void* b;
+        size_t bsize;
+    } buffer; //!< @protected Backend-Related buffer
```
neithernut: Having an additional buffer here doesn't make any sense. We can work on the corresponding `ws_connbuf` in the connector directly.

matthiasbeyer:
> Introducing the possibility to perform serialisation and deserialisation in different threads is a good idea. > However, I don't see any point in having them both based on a general serializer backend.

The point I made this is because they share the same variables. They both operate on a transaction and a buffer, just in different manner. So it is much more convenient to implement a supertype for them and providing the functionality for setting/getting members than implementing everything two times.

src/serialize/backend.h
matthiasbeyer: True.

src/serialize/backend.h
```diff
@@ -0,0 +1,112 @@
...
+};
```
neithernut, line 51: How is this supposed to work, anyways? You didn't even introduce callbacks.

src/serialize/backend.h
matthiasbeyer, line 47: The advantages for having a `ws_object` here are clear: You can do locking without having to implement it again as well as you have a runnable object here, so you can simply call `ws_object_run(serializer)` here. The advantages outperform the disadvantages (which are only overhead, or is there something else) here.

src/command/transaction.h
neithernut: It sais `ws_array`, which implied it holds objects.

src/serialize/backend.h
matthiasbeyer, line 51: Not by now. The backend type wouldn't introduce callbacks, though. The callbacks are only set in the actual implementation, which are not covered yet at all. For example in `struct ws_serializer_serialize_backend_json` or something.

The `struct ws_serializer_backend` is a abstract/virtual type.

src/command/transaction.h
matthiasbeyer: Narg, I misread, I'm sorry. ("wl" vs. "ws", you know?)

src/serialize/backend.h
neithernut, line 47: Simply nope.

First thing is you introduce a transaction getter _and_ a transaction setter, when you really need only one of them at the same time. Second: we defined the interface of the serializer to do the following:
- You lik it to a connection buffer
- In the libev handler callback, you try get a _message_. If a _message_ could be extracted, it returns that message. If data is missing, it returns NULL.

This makes callback design much more simple. Eg. you don't have any need for message buffers, queues or any other stuff like that.

Secondly: the code duplication is minimal, as the serializer and the deserializer each have a callback for exactly one task, and, maybe, a pointer to the `ws_connbuf` it's associated with.
That's minimal code-duplication.

matthiasbeyer:
Somehow, the latest fix compiles. But `struct ws_connbuf` is not defined in the source, so why the hell does it compile?

neithernut:
If you define a pointer to a struct, the struct is implicitely forward declarated.
And as soon as you have a pointer, you don't need the size of the struct for compilation.

src/serialize/backend.h
matthiasbeyer, line 47: The getter _and_ setters were added because of the idea of doing `ws_object_run()` on the object. The dereived classes do something, either transforming a transaction to a buffer or the other way round. So If you put a buffer into the backend, you can run it and get a transaction from it as well as the other way round.

That was my idea when writing this.

If I understand things right, you want me to implement an extended object type table which holds the callbacks for what exactly? You need _either_ to pass a buffer _or_ a transaction and get the other one out of the backend. How would you implement an interface for this?

(your patches to this are welcome)

matthiasbeyer:
Thanks for the explanation. But shouldn't that fail when linking?

src/serialize/backend.h
neithernut, line 47: At this time, I'll think I'll open my own PR.
That is, _after_ I've done the things important for reaching the _current_ milestone.

neithernut:
Only if you deref the struct.

matthiasbeyer:
:angry:

Issue #158, opened 2014-10-09
Add surface creation 


neithernut:
The wayland compositor has a callback which let's you create surfaces. Those surfaces should be assigned to a monitor. ATM this may be done statically, in the compositor module.

neithernut:
Blocks #115.

PR #166, opened 2014-10-11
Add surface creation

neithernut:
This PR introduces a preliminary method of creating a surface and attaching it to each monitor present.

### DIFF
--- a/src/compositor/wayland_compositor.c
+++ b/src/compositor/wayland_compositor.c
@@ -31,6 +31,9 @@
 #include <wayland-server.h>
 #include <wayland-server-protocol.h>
 
+#include "compositor/internal_context.h"
+#include "compositor/monitor.h"
+#include "compositor/surface.h"
 #include "compositor/wayland_compositor.h"
 #include "objects/wayland_obj.h"
 #include "util/arithmetical.h"
@@ -85,6 +88,14 @@ bind_compositor(
     void* data, //!< userdata (not used)
     uint32_t version, //!< interface version
     uint32_t serial //!< serial to give the compositor
+);
+
+/**
+ * Add the surface to a monitor
+ */
+static int add_surface_to_monitor(
+    void* surface_, //!< pointer to the surface
+    void const* monitor_ //!< pointer to the monitor
 );
 
 /*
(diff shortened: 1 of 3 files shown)


neithernut:
Note that, at this point, we may test this solution (the acutal communication with the `flowers` program).
I'm not quite sure if I do this today, but I'll notify you when I did it.

src/compositor/wayland_compositor.c
```diff
@@ -156,7 +167,19 @@ create_surface(
...
     uint32_t serial
 ) {
-    //!< @todo: implement
+    struct ws_surface* surface = ws_surface_new(client, serial);
+    if (!surface) {
+        //!< @todo: throw an error by emitting a signal
+        return;
+    }
+
+    // TEMOORARY SOLUTION: put surface on all of the availible monitors
```
matthiasbeyer: Typo.

neithernut:
Targets #158.

src/compositor/wayland_compositor.c
```diff
@@ -187,3 +210,28 @@ bind_compositor(
     wl_resource_set_implementation(resource, &interface, NULL, NULL);
 }
 
+static int add_surface_to_monitor(
+    void* surface_,
+    void const* monitor_
```
matthiasbeyer, line 215: Also here: Are the trailing underscores wanted?

matthiasbeyer, line 215: Btw: Now I know why you did it, because of the casting. But wouldn't it be more convenient to use `s` and `m` as parameter names instead?

neithernut, line 215: I don't see any benefits.

neithernut:
I just ran a test with `weston-flowers`. It said it was unable to initialize an EGL context and that it was using the shm interface. That seemed to work, so this seems to work so far and can now be merged.

However, it also said it needed a xdg shell. See issue #172.

matthiasbeyer:
How exactly did you run the test with weston-flowers? (so we can reproduce)

neithernut:
```
#!/bin/sh

src/waysome&
sleep 1
weston-flower&

sleep 10
killall weston-flower
killall waysome
```

Started the script from the project's root directory.

TheNeikos:
Do you also get a segfault?

matthiasbeyer:
ACK, @neithernut Is this ready?

TheNeikos:
Okay! Disregard my last comment. Was running on the wrong branch.

Looking good! Great job :+1:

PR #168, opened 2014-10-11
Add string value type

xVin:
targets #35

### DIFF
--- a/src/values/string.c
+++ b/src/values/string.c
@@ -89,8 +89,16 @@ ws_value_string_get_str(
 void
 ws_value_string_set_str(
     struct ws_value_string* self,
-    struct ws_string str
+    struct ws_string* str
 ) {
+    if (self && str) {
+        struct ws_string* new_str = getref(str);
+
+        if (new_str) {
+            ws_object_unref(&self->str->obj);           
+            self->str = new_str;
+        }
+    }
 }
 
 /*
--- a/src/values/string.c
+++ b/src/values/string.c
@@ -38,7 +38,7 @@
  *
  */
 static void
-ws_value_string_deinit(
+value_string_deinit(
     struct ws_value* self
 )
 __ws_nonnull__(1)
@@ -52,7 +52,7 @@ ws_value_string_init(
         ws_value_init(&self->val);
 
         self->val.type = WS_VALUE_TYPE_STRING;
-        self->val.deinit_callback = ws_value_string_deinit;
+        self->val.deinit_callback = value_string_deinit;
 
         struct ws_string* str = ws_string_new();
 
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
This looks good and you have my ACK for going on and adding the next interface methods.

matthiasbeyer:
(Kindof) Blocks #91

neithernut:
Looks good.

xVin:
Please review.

src/values/string.c
```diff
@@ -1,5 +1,4 @@
-/*
- * waysome - wayland based window manager
+ /* waysome - wayland based window manager
```
neithernut: Please remove the whitespace before the `/*`.

src/values/string.c
```diff
@@ -25,5 +25,96 @@
...
+#include "objects/string.h"
+#include "values/value.h"
+
+/*
+ *
+ * forward delarations
+ *
+ */
+static void
+ws_value_string_deinit(
```
matthiasbeyer: Please change the name of the function. The `ws_` prefix is only for the public functions, so call it `string_deinit` or something like that!

matthiasbeyer:
Besides the one annotation, this looks good to me.

src/values/string.c
```diff
@@ -25,5 +25,96 @@
...
+}
+
+void
+ws_value_string_set_str(
+    struct ws_value_string* self,
+    struct ws_string* str
+) {
+    if (self && str) {
+        struct ws_string* new_str;
+        new_str = ws_string_dupl(str);
```
matthiasbeyer: Missed this one:

We only need to get a ref on the string object here and then set the pointer.

src/values/string.c
```diff
@@ -25,5 +25,91 @@
...
+    return NULL;
+}
+
+void
+ws_value_string_set_str(
+    struct ws_value_string* self,
+    struct ws_string* str
+) {
+    if (self && str) {
+        self->str = str;
```
matthiasbeyer: There is still the `ws_object_getref()` call missing here.

src/values/string.c
```diff
@@ -25,5 +25,96 @@
...
+}
+
+void
+ws_value_string_set_str(
+    struct ws_value_string* self,
+    struct ws_string* str
+) {
+    if (self && str) {
+        struct ws_string* new_str = (struct ws_string*) 
+                                     ws_object_getref((struct ws_object*) str);
```
neithernut: there is the shiny new `getref` macro, which does the casting for you. You can simply write:

``` C
struct ws_string* new_str = getref(str);
```

---

pro-tipp: If you run into situations like this one, you can also write:

``` C
very_long_typename_t foo;
foo = somefunc();
```

xVin: I wrote it like this at first, but changed it because I wasn't sure if it fit into our style (x__x)

Well, the macro is neater anyway :3

src/values/string.c
```diff
@@ -25,5 +25,95 @@
...
+    struct ws_string* str
+) {
+    if (self && str) {
+        struct ws_string* new_str = getref(str);
+
+        if (new_str) {
+            self->str = new_str;
+        }
+    }
+}
```
neithernut, line 102: You have to unref the old string.

neithernut:
ACK.

Issue #181, opened 2014-10-12
Add testing: ws_set type


matthiasbeyer:
The set type must be tested.

PR #189, opened 2014-10-12
Add set testing

matthiasbeyer:
This PR adds set testing.

Targets #181

### DIFF
--- a/tests/check/objects/ws_set/test.c
+++ b/tests/check/objects/ws_set/test.c
@@ -43,6 +43,51 @@
 #include "tests.h"
 #include "objects/object.h"
 #include "objects/set.h"
+#include "util/arithmetical.h"
+
+/*
+ *
+ * We need a type with a non-ws_object-compare functionality, to be able to test
+ * the set appropriately
+ *
+ */
+
+struct ws_set_test_obj {
+    struct ws_object obj;
+};
+
+static int
+compare_set_test_objs(
+    struct ws_object const* o1,
+    struct ws_object const* o2
+) {
+    return signum(o2 - o1);
+}
+
+ws_object_type_id TEST_ID = {
+    .supertype  = &WS_OBJECT_TYPE_ID_OBJECT,
+    .typestr    = "ws_set_test_obj",
+
+    .init_callback = NULL,
+    .deinit_callback = NULL,
+    .dump_callback = NULL,
+    .run_callback = NULL,
+    .hash_callback = NULL,
+    .cmp_callback = compare_set_test_objs,
+};
+
+struct ws_set_test_obj*
+new_test_obj(void)
+{
+    struct ws_set_test_obj* o = calloc(1, sizeof(*o));
+
+    if (o) {
+        ws_object_init(&o->obj);
+        o->obj.id = &TEST_ID;
+    }
+
+    return o;
+}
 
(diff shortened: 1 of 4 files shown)


matthiasbeyer:
This does not build. Can someone figure out why? I do not understand the error message here.

neithernut:
The compiler doesn't get that the constant number you use for initializing the array is, indeed, constant.
Use a definition and you're fine.

matthiasbeyer:
@neithernut Can you tell me why the build fails (it doesn't find the executeable somehow)?

matthiasbeyer:
@neithernut I'm ready here. The PR includes fixes for the project sources as well, not only testing!

@TheNeikos review as well, please!

tests/check/objects/ws_set/test.c
```diff
@@ -0,0 +1,435 @@
...
+
+static int
+compare_set_test_objs(
+    struct ws_object const* o1,
+    struct ws_object const* o2
+) {
+    if (o1 == o2) {
+        return 0;
+    } else {
+        return ((((void *) o1) > ((void *) o2)) ? -1 : 1);
```
TheNeikos: you could use the signum macro instead of the if construct

matthiasbeyer: Please provide an example for this, I can't wrap my slimy head around it atm.

neithernut:
``` C
return signum(o2 - o1);
```

The macro is part of `util/arithmetical.h`

tests/check/objects/ws_set/test.c
```diff
@@ -0,0 +1,436 @@
...
+static int
+compare_set_test_objs(
+    struct ws_object const* o1,
+    struct ws_object const* o2
+) {
+    if (o1 == o2) {
+        return 0;
+    } else {
+        return signum(o2 - o1);
+    }
```
neithernut: You don't need the `if` at all. `signum(o2 - o1)` yields `0` if `o2 == o1`.

tests/check/objects/ws_set/test.c
```diff
@@ -0,0 +1,436 @@
...
+    ck_assert(set != NULL);
+}
+
+static void
+test_set_setup_objs(void)
+{
+    test_set_setup();
+    int i;
+
+    for (i = N_TEST_OBJS - 1; i; --i) {
```
neithernut: Either declare `i` inside the `for` or make it a `while`.

tests/check/objects/ws_set/test.c
```diff
@@ -0,0 +1,436 @@
...
+    ck_assert(set_b == NULL);
+
+    set_a = ws_set_new();
+    set_b = ws_set_new();
+
+    ck_assert(set_a != NULL);
+    ck_assert(set_b != NULL);
+
+    int i;
+    for (i = N_TEST_OBJS - 1; i; --i) {
```
neithernut: Same as above.

matthiasbeyer:
@neithernut , @TheNeikos please review, this should go into the sources asap, as fixes are included here.

neithernut:
ACK for squashing.

matthiasbeyer:
No need to.

neithernut:
Then ACK from me.

Issue #188, opened 2014-10-12
Add command representation and interface


neithernut:
We'll introduce a basic representation for commands and command implementation which will be used to construct and process transactions.

neithernut:
Blocks #190.

PR #192, opened 2014-10-13
Add command infrastructure

neithernut:
This PR introduces an initial version of the command infrasructure.
Targets #188.

Please let me know if and how the documentation is not sufficient.

### DIFF
--- a/src/command/command.c
+++ b/src/command/command.c
@@ -27,3 +27,18 @@
 
 #include "command/command.h"
 
+
+/*
+ *
+ * Interface implementation
+ *
+ */
+
+struct ws_command*
+ws_command_get(
+    char const* name //!< name of the command
+) {
+    //!< @todo: implement
+    return NULL;
+}
+
--- a/src/command/command.c
+++ b/src/command/command.c
@@ -1,0 +1,29 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "command/command.h"
+
--- a/src/command/command.h
+++ b/src/command/command.h
@@ -99,5 +99,15 @@ struct ws_command {
     } func; //!< @public function callback
 };
 
+/**
+ * Statement
+ *
+ * This datatype represents a statement, which is part of an transaction
+ */
+struct ws_statement {
+    struct ws_command const* command; //!< @public command to invoke
+    struct ws_command_args args; //!< @public arguments to invoke the command with
+};
+
 #endif // __WS_COMMAND_COMMAND_H__
 
--- a/src/command/command.h
+++ b/src/command/command.h
@@ -28,5 +28,38 @@
 #ifndef __WS_COMMAND_COMMAND_H__
 #define __WS_COMMAND_COMMAND_H__
 
+#include <malloc.h>
+
+// forward declarations
+struct ws_value;
+
+/**
+ * A command argument
+ *
+ * There are two types of arguments: the ones passed directly and the ones which
+ * are already on the stack.
+ */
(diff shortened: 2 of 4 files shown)


src/command/command.h
```diff
@@ -0,0 +1,112 @@
...
+/**
+ * Command type
+ *
+ * This datatype represents a command, which may be invoked in a transaction.
+ */
+struct ws_command {
+    char const* name; //!< @public name of the command
+
+    enum {
+        regular, //!< it's a regulat command
```
TheNeikos: `regular`

src/command/command.h
```diff
@@ -0,0 +1,112 @@
...
+ * There are two types of arguments: the ones passed directly and the ones which
+ * are already on the stack.
+ */
+struct argument {
+    enum {
+        indirect, //!< argument is given by position
+        direct //!< argument is given directly
+    } type; //!< @public type of argument (direct or indirect)
+    union {
+        ssize_t pos; //!< a position
```
matthiasbeyer: Why `ssize_t` - documentation missing here.

src/command/command.h
```diff
@@ -0,0 +1,112 @@
...
+ */
+struct ws_command_args {
+    size_t num; //!< @public number of arguments
+    struct argument* vals; //!< @public arguments
+};
+
+/**
+ * Function type for command implementation
+ *
+ * The function
```
matthiasbeyer: Here's documentation missing.

src/command/command.h
```diff
@@ -0,0 +1,112 @@
...
+typedef int (*ws_special_command_func)(struct ws_command_processor*,
+                                       struct ws_command_args*);
+
+/**
+ * Command type
+ *
+ * This datatype represents a command, which may be invoked in a transaction.
+ */
+struct ws_command {
+    char const* name; //!< @public name of the command
```
matthiasbeyer: Could this possibly be a `char const* const name`?

src/command/command.h
```diff
@@ -0,0 +1,112 @@
...
+
+    enum {
+        regular, //!< it's a regular command
+        special //!< it's a special command (like jump)
+    } command_type; //!< @public type of the command
+
+    union {
+        ws_regular_command_func regular; //!< @public regular commands' callback
+        ws_special_command_func special; //!< @public special commands' callback
+    } func; //!< @public function callback
```
matthiasbeyer, line 99: Suggestion: The command type and the callback are directly related. Why not putting it into an own struct like so:

``` C
struct {
    enum { /* */ } type;
    union { /* */ } func;
} callback;
```

neithernut, line 99: If you look at the function definition, you'll notice that there is a lot of documentation, which I didn't want in the struct.

matthiasbeyer, line 99: That's not what I wanted to explain:

You actually can put the union and the enum above into an own scope, as they belong to eachother very closely.

neithernut, line 99: Suboptimal alignment.

matthiasbeyer, line 99: Are you sure about this? The additional struct doesn't add any data to the type, just a scope.

neithernut, line 99: The struct may align to `2*sizeof(void*)`.

src/command/command.h
neithernut: I would do this in a more central place. Preferably in the module description.

matthiasbeyer: I'd prefer a note here. Documentation belongs to the code it documents, not somewhere else!

neithernut: The question is whether you get it, if I document it here.

matthiasbeyer: That's true. Then, a `@see` would be good.

neithernut:
Any more anotations on this one?

src/command/command.h
```diff
@@ -0,0 +1,113 @@
...
+
+#include <malloc.h>
+
+// forward declarations
+struct ws_command_processor;
+struct ws_value;
+
+/**
+ * A command agument
+ *
```
xVin, line 39: typo: argument

src/command/command.h
```diff
@@ -0,0 +1,113 @@
...
+struct ws_command_processor;
+struct ws_value;
+
+/**
+ * A command argument
+ *
+ * There are two types of arguments: the ones passed directly and the ones which
+ * are already on the stack.
+ */
+struct argument {
```
matthiasbeyer: This does not fit our naming scheme here.

neithernut: Uh, forgot to prefix it when I pulled it out of another struct.

src/command/command.c
```diff
@@ -0,0 +1,44 @@
...
+
+
+/*
+ *
+ * Interface implementation
+ *
+ */
+
+struct ws_command*
+ws_command_get(
```
matthiasbeyer, line 38: Documentation missing. I don't know what this function is meant to be, but I guess that "get" is a bit inappropriate here, but I don't know what this function is for...

neithernut, line 38: The function is documented in the header.

src/command/command.h
```diff
@@ -0,0 +1,124 @@
...
+
+/**
+ * Command arguments
+ * 
+ * This struct represents command arguments applied to a command as part of a
+ * statement.
+ */
+struct ws_command_args {
+    size_t num; //!< @public number of arguments
+    struct argument* vals; //!< @public arguments
```
matthiasbeyer: Mustn't this be renamed to `ws_argument`?

neithernut: Er... yep.

src/command/command.c
matthiasbeyer, line 38: Whoops.

neithernut:
Any more annotations? I'm currently waiting for this to be merged.

matthiasbeyer:
Nope, no annotations from me. ACK.

PR #193, opened 2014-10-13
Added id to monitor

TheNeikos:
This adds an id to the monitor struct as it simplifies a few things down the line.


neithernut:
ACK.

matthiasbeyer:
ACK.

PR #196, opened 2014-10-13
Fixed ws_set_get selecting the opposite

TheNeikos:
This PR fixes a bug in ws_set_get.

Libreset expects a boolean return value, but we give it a POSIX style return value. This inverts the logic.


neithernut:
ACK.

Issue #194, opened 2014-10-13
Add string type setter function


matthiasbeyer:
We need something like `ws_string_set_from_raw()` for the string type.

Blocks #91 .

xVin:
Ok I'm on it.

xVin:
Can be closed.

PR #198, opened 2014-10-13
Add ws_string_set_from_raw()

xVin:
targets #194

### DIFF
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -294,6 +294,42 @@ ws_string_raw(
     return output;   
 }
 
+void
+ws_string_set_from_raw(
+    struct ws_string* self,
+    char* raw
+){  
+    if (!self) {
+        return;
+    }
+    
+    int32_t len = strlen(raw);
+    UChar* conv_raw = calloc(len + 1, sizeof(*conv_raw));
+        
+    if (!conv_raw) {
+        return;
+     }
+    
+    int32_t dest_len;
+    UErrorCode err;
+
+    conv_raw = u_strFromUTF8(conv_raw, len + 1, &dest_len, raw, len, &err);
+        
+    if (U_FAILURE(err)) {
+        return;       
+    }
+
+    ws_object_lock_write(&self->obj);
+
+    free(self->str);
+        
+    self->str = conv_raw;
+    self->is_utf8 = true;
+    self->charcount = len;
+
+    ws_object_unlock(&self->obj);   
+}
+
 /*
  *
  * Static Function Implementations
--- a/src/objects/string.h
+++ b/src/objects/string.h
@@ -193,6 +193,18 @@ ws_string_raw(
     struct ws_string* self
 );
 
+/**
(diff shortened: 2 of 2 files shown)


xVin:
Kindly review. 

(I dunno why this PR has no compiler check yet o.o)

matthiasbeyer:
@xVin look here: http://ci.neikos.me/github.com/waysome/waysome

src/objects/string.c
```diff
@@ -294,6 +294,42 @@ ws_string_raw(
     return output;   
 }
 
+void
+ws_string_set_from_raw(
+    struct ws_string* self,
+    char* raw
+){  
+    if (self) {
```
neithernut: I think it would be cleaner to do an early return here.

src/objects/string.c
```diff
@@ -294,6 +294,44 @@ ws_string_raw(
...
+
+    conv_raw = u_strFromUTF8(conv_raw, len + 1, &dest_len, raw, len, &err);
+        
+    if (U_FAILURE(err)) {
+        return;       
+    }
+
+    ws_object_lock_write(&self->obj);
+
+    self->str = realloc(self->str, (len + 1) * sizeof(*self->str));
```
neithernut: I do not get why you do the realloc here. Shouldn't `self->str` simply be `free()`d and set to `conv_raw`?

xVin: True.

src/objects/string.c
```diff
@@ -294,6 +294,44 @@ ws_string_raw(
...
+    UChar* conv_raw = calloc(len + 1, sizeof(*conv_raw));
+        
+    if (!conv_raw) {
+        return;
+     }
+    
+    int32_t dest_len;
+    UErrorCode err;
+
+    conv_raw = u_strFromUTF8(conv_raw, len + 1, &dest_len, raw, len, &err);
```
matthiasbeyer, line 316: I don't get why the function is called `strFromIUTF8` if it does (or is intended to do) `UTF8FromStr` - are you sure this is correct here?

xVin, line 316: This funtion generates a UChar string from an UTF-8 string (= char*), so it should be what we want here.

src/objects/string.c
```diff
@@ -294,6 +294,44 @@ ws_string_raw(
...
+        
+    if (U_FAILURE(err)) {
+        return;       
+    }
+
+    ws_object_lock_write(&self->obj);
+
+    free(self->str);
+        
+    if (self->str) {
```
matthiasbeyer: This is misleading: You do a `free()` beforehand, you should do a `self->str = NULL` there, too. This if would then result in beeing `false` always.

I guess the `if` must be removed.

xVin: Oops, forgot that in the previous fix. Will be fixed in a jiffy :o

src/objects/string.c
neithernut, line 316: Yep.

neithernut:
ACK.

PR #199, opened 2014-10-13
Moved framebuffer creation out of monitor

TheNeikos:
This moves the framebuffer creation out of the monitor.

Thus we can create framebuffers on the fly.

### DIFF
--- a/src/compositor/frame_buffer.c
+++ b/src/compositor/frame_buffer.c
@@ -1,0 +1,189 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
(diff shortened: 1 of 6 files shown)


src/compositor/frame_buffer.c
```diff
@@ -0,0 +1,189 @@
...
+ *
+ */
+
+static bool
+frame_buffer_deinit(
+    struct ws_object* obj
+) {
+    struct ws_frame_buffer* self = (struct ws_frame_buffer*) obj;
+    if (self->obj.buffer) {
+        munmap(self->obj.buffer, self->obj.size);
```
matthiasbeyer, line 174: Are there no supertype methods for this? Would be much cleaner, I guess. - But this could be done in a seperate PR, just open an issue for it, please.

matthiasbeyer:
Looks good for me so far, but I guess we should do a refactoring of the buffer types.
I also suggest for adding a subdirectory `src/compositor/buffer/` or something for the buffer types. At least, the `src/compositor/` directory gets really full and I would like to refactor the source to be a bit more structured. This can be done in a seperate PR, of course. I just want to mention it and if you guys like the idea of a refactoring for the compositor, please open an issue for it!

neithernut:
Should not be merged before #202 and #203 are merged.

src/compositor/frame_buffer.c
matthiasbeyer, line 174: @TheNeikos explained it to me and it sounds reasonable to use the variable directly here  (You don't know if the getter modifies the buffer before returning it).

Anyways, I'm not happy with that.

matthiasbeyer:
Can you add a space after casts the next time: `(void*) foo` instead of `(void*)foo` ? I'm not bitching around this time, but I will next time. It's a readability thing, you know?

matthiasbeyer:
ACK btw.

PR #200, opened 2014-10-13
Fix iterations up the inheritance tree in objects

neithernut:
This fixes A bug I introduced into the code while fixing other stuff.

### DIFF
--- a/src/objects/object.c
+++ b/src/objects/object.c
@@ -209,6 +209,8 @@ ws_object_run(
         if (type == &WS_OBJECT_TYPE_ID_OBJECT) {
             return false;
         }
+
+        type = type->supertype;
     }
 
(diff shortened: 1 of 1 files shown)


neithernut:
Self ACK for the sake of progress.

matthiasbeyer:
There was already a fix in one of my branches, whatthehell?

matthiasbeyer:
And why the hell are you doing self-ACKs? This is not what our workflow is! Stop this!

TheNeikos:
We needed it at that _moment_. I was beside him as he wrote it, so it was not a complete "self ACK"

matthiasbeyer:
And here we have the point why I want the actual "ACK" in the issue.

PR #203, opened 2014-10-13
Add compositor wl shell

neithernut:
This PR adds a preliminary implementation of the `wl_shell`.

### DIFF
--- a/src/compositor/wayland_shell_surface.c
+++ b/src/compositor/wayland_shell_surface.c
@@ -1,0 +1,46 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "compositor/wayland_shell_surface.h"
+
+/*
+ *
+ * Interface implementation
+ *
+ */
+
+ws_object_type_id WS_OBJECT_TYPE_ID_SHELL_SURFACE = {
+    .supertype  = &WS_OBJECT_TYPE_ID_WAYLAND_OBJ,
+    .typestr    = "ws_shell_surface",
+
+    .deinit_callback    = NULL,
+    .init_callback      = NULL,
+    .dump_callback      = NULL,
+    .run_callback       = NULL,
+    .hash_callback      = NULL,
+};
+
--- a/src/compositor/wayland_shell_surface.c
+++ b/src/compositor/wayland_shell_surface.c
@@ -25,6 +25,7 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <malloc.h>
 #include <wayland-server.h>
 
 #include "compositor/wayland_shell_surface.h"
@@ -32,7 +33,7 @@
 /**
  * Version of the wayland surface interface we're implementing
  */
-#define WAYLAND_SURFACE_VERSION (1)
+#define WAYLAND_SHELL_SURFACE_VERSION (1)
 
 /*
  *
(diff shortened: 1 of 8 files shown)


src/compositor/wayland_shell_surface.c
```diff
@@ -0,0 +1,372 @@
...
+    .hash_callback      = NULL,
+};
+
+struct ws_shell_surface*
+ws_shell_surface_new(
+    struct wl_client* client,
+    struct ws_surface* surface,
+    uint32_t serial
+) {
+    struct ws_shell_surface* self = calloc(1, sizeof(struct ws_shell_surface));
```
matthiasbeyer: `sizeof(*self)` here, please.

src/compositor/wayland_shell_surface.c
```diff
@@ -0,0 +1,372 @@
...
+    struct wl_resource* resource;
+    resource = wl_resource_create(client, &wl_shell_surface_interface,
+                                  WAYLAND_SHELL_SURFACE_VERSION, serial);
+    if (!resource) {
+        goto cleanup_surface;
+    }
+
+    // set the implementation
+    wl_resource_set_implementation(resource, &interface,
+                                   ws_object_getref(&self->wl_obj.obj),
```
matthiasbeyer: `ws_object_getref()` returns `NULL` on failure, you know that?

src/compositor/wayland_shell_surface.c
```diff
@@ -0,0 +1,372 @@
...
+                                   resource_destroy);
+
+    // finish the initialization
+    ws_wayland_obj_init(&self->wl_obj, resource);
+    self->wl_obj.obj.id = &WS_OBJECT_TYPE_ID_SHELL_SURFACE;
+
+    // initialize the members
+    self->surface = getref(surface);
```
matthiasbeyer, line 243: Why are you using the function call above but the macro here?

src/compositor/wayland_shell_surface.h
```diff
@@ -0,0 +1,65 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef __WS_WL_SHELL_SURFACE_H__
+#define __WS_WL_SHELL_SURFACE_H__
+
+#include "objects/wayland_obj.h"
+
+// forward declarations
```
matthiasbeyer, line 33: Why?

src/compositor/wayland_shell_surface.c
neithernut, line 243: Because the function takes a `void*`, so the type is irrelevant.

src/compositor/wayland_shell_surface.h
neithernut, line 33: Because this header doesn't depend on a bunch of other headers this way. This speeds up differential builds a lot.

matthiasbeyer, line 33: This sounds really back-practice-like to me! I'd like to hear @TheNeikos and @manuelmessner on this!

neithernut, line 33: Maybe to _you_ it sounds like a bad practice, but there's absolutely _no point_ in pulling in >100 lines of header code if you really only need a struct declration you can use for a pointer.
Also, it prevents you from e.g. accidentally declaring functions taking structs (not pointer to structs) as parameters, fucking up the interface.

matthiasbeyer, line 33: Maybe our architecture sucks if we need to pull in ">100 lines of header code" for just one type...

:laughing:

neithernut, line 33: Nope, we need > 100 lines of header code because we have a 27-line-license-header, we need 5 lines for include guards and at least 7 lines (most of the time it's more than 10) for each function declaration.
And I'm not even talking about doxygen groups here.

Plus, if you include a lot of headers in one header, how many lines of header code do you think you end up pulling in? Most of the headers you include include headers themselfes. Think of _that_, too.

matthiasbeyer, line 33: Yes, and if the architecture would be a bit more distributed in manner of type declarations, we may would end up pulling in less lines here. I do not say "your code sucks" here, but I say "our code _architecture_ sucks" here.

neithernut, line 33: Nope, the architecture doesn't suck that much. All our modules are pretty much self contained. The only thing that will be pulled in from a lot of modules is in `util`, `objects` and `values`.

However, you _can_ make thinks cleaner by using forward declarations. I pay attention to header's footprints, even if they are small. That's why I do forward declarations.

If you would kindly ACK now, or at least, come up with some other requests?

matthiasbeyer, line 33: I'm undecided here. I would like to wait for comments from @TheNeikos and if he ACKs, I'm fine with it.

Are you fine with this suggestion? I'll notify him directly right now.

neithernut, line 33: I am.

TheNeikos, line 33: The only thing I don't like is that it looks like tight coupling. _However_ this is part of the compositor module! So this is kind of expected. I have no trouble with this, however I would like a comment specifying which file originally holds the definition of the _real_ struct.

matthiasbeyer, line 33: Sounds like a good agreement for me.

neithernut, line 33: If you look at it in another way, it's _less_ coupled, as you only depend on the type to be present, but not on the internals.
Anyways, I added the note.

TheNeikos, line 33: Awesome!

manuelmessner, line 33: > Anyways, I added the note.

ACK

matthiasbeyer:
I assume you did not change any further things here except the documentation?

TheNeikos:
ACK

PR #204, opened 2014-10-14
Use "__typeof__" instead of "typeof" for getref() macro

neithernut:
This is a very simple fix meant to prevent compiler warnings.
Still uses gcc/clang extensions...


matthiasbeyer:
ACK.

PR #206, opened 2014-10-14
Add logging calls in ws_object implementation

matthiasbeyer:
This PR adds a few calls to ws_log() in the object type
implementation.


TheNeikos:
ACK

PR #208, opened 2014-10-15
Fix: Add space after debug prefix for object type

matthiasbeyer:


neithernut:
We could also adjust the log message creation so it adds the whitespace and the brackets.

matthiasbeyer:
But I would introduce another feature for that, something like a member `tag`. So you set `tag` to `"Object"` and it creates `"[Object] "` out of it. And the tag should then go _before_ the actual prefix.

neithernut:
That doesn't sound sane.

matthiasbeyer:
Why not, that's what you suggested. A prefix is not the same as a `"[something]"` tag.

neithernut:
Having a prefix _and_ a tag sounds nonsensical.

matthiasbeyer:
I consider a tag as some meta-information, where a log is located (Hey, I'm from the connector implementation) and a prefix for in what context it happened (Hey, I'm logging here while receiving stuff).

Another idea would be, to have a list of tags which gets prepended to the actual log. But that results in a way more complex algorithm that way, and as we may replace the whole logging stuff at some point (maybe) it sounds like unrequired effort for me.

matthiasbeyer:
I guess the discussion should be held in an appropriate issue, not here. Can you please review and ack?

matthiasbeyer:
Can we please get this merged or not?

neithernut:
ACK, albeit the underlying issue stays unresolved.

PR #209, opened 2014-10-15
Add command list

neithernut:
This PR introduces template files for the the command list and a build environment setup to acutally generate it. It takes care of the list being alphabetically ordered.

One thing I'm still unsure about is whether I should add a source file to the `SOURCE_FILES` for each header generated. It would mean that you don't have to do it yourself, but it would also _force_ you to have those sources.

### DIFF
--- a/src/command/CMakeLists.txt
+++ b/src/command/CMakeLists.txt
@@ -21,6 +21,7 @@
 # List of command files
 #
 set(WS_COMMAND_FILES
+    arithmetical.commands
 )
 
 
--- a/src/command/CMakeLists.txt
+++ b/src/command/CMakeLists.txt
@@ -1,0 +1,13 @@
+
+#
+# List of command files
+#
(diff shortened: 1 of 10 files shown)


neithernut:
Btw: the source doesn't build yet, because I don't have any functions.

src/command/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+#
+# Each line of such a file contains the command's name, a semicolon and the
+# type of the command ("regular" or "special").
+# For each line, a command declaration will be generated and put into a header
+# with the same basename as the command name.
+# The command functions declarated in the generated header should be implemented
+# in the corresponding source file.
+#
+# All the commands will be added to the generated command list in list.c,
+# making them globally availible through the ws_command_get() method.
```
matthiasbeyer: Typo.

matthiasbeyer:
ACK.

Issue #197, opened 2014-10-13
Find a default cursor


neithernut:
We should ship a default cursor, in PNG format.
We either have to make our own or find one. There are websites which give you tons of free icons. You have to look out for compatible licensing, however.

xVin:
Would you prefer a white or a black default cursor?

matthiasbeyer:
I'm trying to do this right now, but I guess competing approaches would be nice, so go for it.

I'd like to suggest to ship both, a white-black and a black-white one!

neithernut:
You could easily invert the cursor, so it doesn't really matter if it's black or white.

matthiasbeyer:
![BLACK OR WHITE; YEAR](http://itunes-music.com/wp-content/uploads/2012/07/michael-performing-black-or-white.jpg?eb8c7d)

matthiasbeyer:
[My approach](http://webuser.hs-furtwangen.de/~beyermat/cursor.png)

neithernut:
I created a smaller variant from your version. Additionally, I added a version with a small effect:
http://webuser.hs-furtwangen.de/~ganzjuli/waysome/cursor/

matthiasbeyer:
Looks good, too. But I don't see the difference besides in size. And I don't see the effect.

xVin:
He means the shading effect.

neithernut:
The shading effect is meant to be minimal. This gives the cursor a little bit of depth without distracting the user.

Issue #179, opened 2014-10-12
Add a cursor


TheNeikos:
For now it would be sufficient if the mouse would be just placed at the top left or some other position.

TheNeikos:
It should read an image on the file system and should be placed _over_ every other type of buffer.

This is where compositing comes into play

neithernut:
Yep. Wasn't there a way to pass a plane to libdrm and marking it as a pointer/arrow? I thought I saw something there...

Note: virtually all modern graphics cards support hardware based pointer drawing, and most people use this.

TheNeikos:
Yup! That's what we are going to use

PR #211, opened 2014-10-15
Add cursor to the compositor

TheNeikos:
This PR adds a sample cursor image as well as a working cursor implementation that is integrated into wayland. (As by spec)


neithernut:
Targets #179 and #197.

TheNeikos:
Okay, this is now open to reviews

src/compositor/cursor.c
```diff
@@ -0,0 +1,174 @@
...
+ws_cursor_new(
+    struct ws_framebuffer_device* dev,
+    struct ws_image_buffer* cur
+) {
+    struct ws_cursor* self = calloc(1, sizeof(*self));
+    ws_object_init((struct ws_object*) self);
+    self->obj.id = &WS_OBJECT_TYPE_ID_CURSOR;
+    self->cur_fb_dev = dev;
+    self->cursor_fb = ws_frame_buffer_new(dev, CURSOR_SIZE, CURSOR_SIZE);
+    self->x_hp = 1;
```
matthiasbeyer, line 83: Can you add notes what these magic constants are for (also the three following lines)?

src/compositor/cursor.c
```diff
@@ -0,0 +1,174 @@
...
+    self->y_hp = 1;
+    self->x = 350;
+    self->y = 350;
+    self->default_cursor = cur;
+    ws_buffer_blit((struct ws_buffer*) self->cursor_fb,
+            (struct ws_buffer*) self->default_cursor);
```
matthiasbeyer: Can you add more indentation here?

src/compositor/cursor.c
```diff
@@ -0,0 +1,174 @@
...
+    return self;
+}
+
+void
+ws_cursor_set_position(
+    struct ws_cursor* self,
+    int x,
+    int y
+) {
+    int w = ws_buffer_width((struct ws_buffer*)self->cur_mon->buffer);
```
matthiasbeyer: Can you add a space after the cast?

src/compositor/cursor.c
```diff
@@ -0,0 +1,174 @@
...
+    int h = ws_buffer_height((struct ws_buffer*)self->cur_mon->buffer);
+
+    self->x = CLAMP(-self->x_hp, x, w);
```
matthiasbeyer, line 105: Why `-self->x_hp` and not the positive value, can you add a comment here?

src/compositor/cursor.c
```diff
@@ -0,0 +1,174 @@
...
+void
+ws_cursor_redraw(
+    struct ws_cursor* self
+) {
+    int w = ws_buffer_width((struct ws_buffer*) self->cursor_fb);
+    int h = ws_buffer_height((struct ws_buffer*) self->cursor_fb);
+
+    if (drmModeSetCursor2(self->cur_fb_dev->fd, self->cur_mon->crtc,
+            self->cursor_fb->handle, w, h,
+            self->x_hp, self->y_hp) != 0) {
```
matthiasbeyer: I would highly prefer a temporary variable here instead of doing the function call right inside the if.

src/compositor/cursor.c
```diff
@@ -0,0 +1,174 @@
...
+        ws_log(&log_ctx, LOG_CRIT, "Could not set cursor");
+        ws_log(&log_ctx, LOG_CRIT, "State was: crtc: %d, handle: %d, "
+                "height: %d, width: %d",
+                self->cur_mon->crtc, self->cursor_fb->handle, w, h);
+    }
+    if (drmModeMoveCursor(self->cur_fb_dev->fd, self->cur_mon->crtc,
```
matthiasbeyer: Here too, I would call the function outside the if.

src/compositor/module.c
```diff
@@ -206,6 +208,20 @@ ws_compositor_init(void) {
         return retval;
     }
 
+    retval = ws_wayland_seat_init();
+    if (retval < 0) {
+        return retval;
+    }
+
+    struct ws_image_buffer* cursor =
+        ws_image_buffer_from_png("share/waysome/cursor.png");
```
matthiasbeyer, line 218: We should add a todo here, that this should be adapted as soon as possible.

src/compositor/module.c
```diff
@@ -273,7 +294,7 @@ set_monitor_modes(
         return 0;
     }
     ws_monitor_set_mode_with_id(monitor,
-            ws_set_cardinality(&monitor->modes) - 1);
+            0); // Set to the biggest mode
```
matthiasbeyer: Do you need this on a new line here?

neithernut: You can move the comment on the line just before the call.

src/objects/set.h
```diff
@@ -226,4 +226,12 @@ ws_set_select(
     void* proc_etc //!< Additional parameter for the processor function
 );
 
+/**
+ * Selects any elements (which might be the same as previously)
```
matthiasbeyer: Shouldn't it be "Select first possible element" (note the singular)?

src/objects/set.h
```diff
@@ -226,4 +226,12 @@ ws_set_select(
...
+ *
```
matthiasbeyer, line 231: `@return` missing.

neithernut:
ACK.

PR #212, opened 2014-10-15
Fix void cast in wayland_shell.c

neithernut:
This was a typo, which causes clang based builds to fail if `HARD_MODE=ON` (and causes a nasty warning if we are not in hard mode).


TheNeikos:
ACK

neithernut:
Selfrigheous sui-merge.

Issue #160, opened 2014-10-10
Implement the ws_connector


neithernut:
The `ws_connector`, which is btw _not an object_, has to be implemented according to spec.
Additionally to the stuff the spec says, a read-only mode should be added, which simply does not initialize or use the `outbuf`.

The initialisation funtion should take an file descriptor along with the connector itself.

xVin:
Do you think I could try doing this? 
I'd like to give it a shot. :-)

matthiasbeyer:
@xVin : I think you can do the interface, which should be one PR. But not the implementation, this is far to heavy IMHO for you by now.

xVin:
kk, I'll do the interface then. 
As for the implementation: Maybe we could do it pair programming style? 
But if you think it's too far above my head, that's ok...

xVin:
Seeing that this was assigned to manu: Am I still 'allowed' to do the interface?

neithernut:
Yes, of course.

matthiasbeyer:
@xVin How much is done?

xVin:
Expect the PR this evening.

neithernut:
Blocks #219.

xVin:
Can be closed.

PR #217, opened 2014-10-16
Add ws connector 

xVin:
targets #160 

This is what I figured from the specs.
I admit I'm not completely sure how this thing is gonna work (especially the flushing / the actual data transfer via outbuf).
But I wanted to do this in order to gain a better understanding of how stuff works, so I hope that's still alright...

I'm thankful for corrections and/or hints about which prototypes are still missing.


src/connector/connector.h
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
```
matthiasbeyer: Include-guards are missing here.

src/connector/connector.c
```diff
@@ -0,0 +1,26 @@
...
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
```
matthiasbeyer: The include for the header is missing here.

src/connector/connector.h
```diff
@@ -0,0 +1,86 @@
...
+#include <stdbool.h>
+
+#include "connection/connbuf.h"
+
+/**
+ * ws_connector type definition
+ */
+
+struct ws_connector {
+    int fd; //!<file descriptor
```
matthiasbeyer: Please add the visibility annotations in the comment (also for the other members). This is probably always `@protected` here. Also add spaces after the comment markers ("//!< " instead of "//!<").

matthiasbeyer:
I'm not able to tell you much about whether this is sane or not, but I hope the annotations are still helpful!

neithernut:
The interface itself looks good so far. If you like you can proceed implementing the interface as sson as you're done with the missing elements in the header.

---

I'll be in C2.16 most of the time tomorrow (and availible via irc), if you have any questions.

xVin:
Thanks for the comments - I'm at the office atm, but will continue working on this as soon as I get home. I can't come to Fuwa today (my boyfriend needs the car), but I'll be on IRC and I indeed have some questions :>

xVin:
I added the deinit prototype and fixed some stuff, anything still missing/wrong?

src/connection/connector.h
```diff
@@ -0,0 +1,102 @@
...
+#include "connection/connbuf.h"
+
+/**
+ * ws_connector type definition
+ */
+
+struct ws_connector {
+    int fd; //!< @protected file descriptor
+    struct ws_connbuf* inbuf; //!< @public buffer for read data
+    struct ws_connbuf* outbuf; /**!< @public buffer for data
```
neithernut: Both `inbuf` and `outbuf` should be embedded into the struct directly (remove the `*`).

src/connection/connector.c
```diff
@@ -0,0 +1,133 @@
...
+){
+    char* start;
+    int res;
+    int remaining_mem; //amount of free memory in outbuf
+
+    remaining_mem = self->inbuf.size - self->inbuf.data;
+    start = ws_connbuf_reserve(&self->inbuf, remaining_mem);
+
+    res = read(self->fd, start, remaining_mem);
+    if (res == -1) {
```
neithernut: Any value `<0` is malicious, not just `-1`. So it makes sense to have `res < 0` here.
Also: you _have_ to append something once you reserved stuff. You can, however, just append `0` bytes.

src/connection/connector.c
```diff
@@ -0,0 +1,133 @@
...
+){
+    if (self->readonly) {
+        return -1;
+    }
+
+    size_t used_mem = self->outbuf.data;
+    int res;
+
+    res = write(self->fd, self->outbuf.buffer, used_mem);
+    if (res == -1 ) {
```
neithernut: Same as above, you could have `res < 0` here.

xVin:
Please review this. :)

src/connection/connector.c
```diff
@@ -0,0 +1,133 @@
...
+    char* start;
+    int res;
+    size_t remaining_mem; //amount of free memory in outbuf
+
+    remaining_mem = self->inbuf.size - self->inbuf.data;
+    start = ws_connbuf_reserve(&self->inbuf, remaining_mem);
+
+    res = read(self->fd, start, remaining_mem);
+    if (res < 0) {
+        return res;
```
neithernut, line 99: Please call `ws_connbuf_append()` once you reserved memory, so the connbuf is ""unblocked" before they return. (so you can call `ws_connbuf_reserve()` or `ws_connbuf_discard()`)

neithernut:
ACK from me. The history looks clean, too. Are there any changes you want to perform?

---

Looks like there _are_ still changes commitg in.

src/connection/connector.c
```diff
@@ -0,0 +1,134 @@
...
+ws_connector_deinit(
+    struct ws_connector* self
+){
+    ws_connbuf_deinit(&self->inbuf);
+
+    if (!self->readonly) {
+        ws_connbuf_deinit(&self->outbuf);
+    }
+
+    return;
```
matthiasbeyer: return not required here.

matthiasbeyer:
One enhancement for the history: Move the commit which adds the CMakeLists.txt entry right after the commit which adds the `.c` file. You don't have to do this now, but next time.

So far, this looks good to me (besides the one annotation).

xVin:
I had to fix a small error to prevent ws_connector_read() from accidentally returning 0, should be fine now.

src/connection/connector.c
```diff
@@ -0,0 +1,132 @@
...
+    size_t remaining_mem; //amount of free memory in outbuf
+
+    remaining_mem = self->inbuf.size - self->inbuf.data;
+    start = ws_connbuf_reserve(&self->inbuf, remaining_mem);
+
+    res = read(self->fd, start, remaining_mem);
+    if (res < 0) {
+        res = ws_connbuf_append(&self->inbuf, 0); //unblocks the buffer
+        return -1;
+    }
```
neithernut, line 100: I'd propose you ignore the return value of `ws_connbuf_append()` and return `res`, which you get from the `read`. That way you have a chance to catch `read()` returning `-EAGAIN` (which means that no error occured but that there is just no data atm.)

xVin, line 100: I simply didn't know that I can ignore return values like this without even getting a compiler warning... still so much to learn ;_;

neithernut:
ACK.

matthiasbeyer:
ACK.

Issue #191, opened 2014-10-12
Add message type "transaction"


neithernut:
Add a message type which holds a transaction.

matthiasbeyer:
Let me add a short note here (not _that_ closely related): We should have a look at the language definition some time in the near future. The language is already there, but we did not agree on it.

neithernut:
Blocks #218.

matthiasbeyer:
Can be closed, right?

PR #221, opened 2014-10-17
Add transaction type

matthiasbeyer:
Targets #191


matthiasbeyer:
@neithernut Please review calmly!

matthiasbeyer:
@neithernut please review.

src/objects/message/transaction.c
```diff
@@ -0,0 +1,109 @@
...
+        return NULL;
+    }
+
+    if (!ws_message_init(&t->m, id)) {
+        free(t);
+        return NULL;
+    }
+
+    t->cmds = cmds;
+    t->name = name;
```
neithernut: do a `getref()` here.

src/objects/message/transaction.c
```diff
@@ -0,0 +1,109 @@
...
+    t->flags = flags;
```
neithernut, line 54: Reoerder these to match the order of the parameters.

src/objects/message/transaction.c
```diff
@@ -0,0 +1,109 @@
...
+    return t->flags;
+}
+
+struct ws_string*
+ws_transaction_name(
+    struct ws_transaction* t
+) {
+    struct ws_string* s = NULL;
+
+    ws_object_lock_read(&t->m.obj);
```
neithernut: Why do you lock here, but not in the other getters? (the locks should not be neccessary).

matthiasbeyer: Because the other getters are not dangerous, because there are no setters. But for this one, there is an object which could be referenced from other things, too.

neithernut: `getref()` takes cre of this.

src/objects/message/transaction.c
```diff
@@ -0,0 +1,110 @@
...
+    if (!ws_message_init(&t->m, id)) {
+        free(t);
+        return NULL;
+    }
+
+    t->cmds = cmds;
+    if (name) {
+        ws_object_getref(&name->obj);
+    }
+    t->name = name;
```
neithernut: Just do `t->name = getref(name);`. It does the casting and performs a nullcheck.

src/objects/message/transaction.c
```diff
@@ -0,0 +1,110 @@
...
+struct ws_string*
+ws_transaction_name(
+    struct ws_transaction* t
+) {
+    struct ws_string* s = NULL;
+
+    s = t->name;
+    if (s) {
+        ws_object_getref(&s->obj);
+    }
```
neithernut: Just do a `getref()`

neithernut:
ACK.

PR #226, opened 2014-10-17
Remove unneccessary check from ws_connbuf_append

neithernut:
In some cases, you want to append 0 bytes to unblock the buffer.


xVin:
ACK.

TheNeikos:
Wouldn't it make sense to add an unblock method?

neithernut:
Mh... maybe, but `ws_connbuf_append(foo, 0)` _must_ unblock the buffer.

matthiasbeyer:
Sounds reasonable to me.

Issue #233, opened 2014-10-18
[Request] `ws_command` interface enhancements


matthiasbeyer:
For my work on the JSON deserializer, I need some more functionality in the `ws_command` type interface:
- Construct a command object, plain, without any values set
- Setters/Getters for 
  - Command name
  - Command type
- Arguments modification
- A finalizer method for "Hey, here is a command, fill out all values I didn't" or something like that

Also, the getter by name should be chaned to return a new `ws_command` object, so I can modify it. 

I need this ASAP.

neithernut:
I think you don't quite understand how these things work...
Look at the documentation. All these fields are public _for a reason_.
Also: you don't have a sequence of commands, but of `ws_stantement`s.
And "Fill out all values I didn't" is plain bullshit.

neithernut:
Concering arguments: have a `ws_array`, collect all the arguments and then memcpy them into the statement.
Same goes for the statements.

matthiasbeyer:
What is that `ws_statement` you're talking of? I don't know that type.

neithernut:
Look at `src/command/command.h`

neithernut:
I'll assume this can be closed now...

matthiasbeyer:
ACK.

PR #234, opened 2014-10-18
Fix ws_transaction_command_list

neithernut:
Should be obvious. We want statements, not commands. Statements are command invocations.
Targets #233.

### DIFF
--- a/src/objects/message/transaction.h
+++ b/src/objects/message/transaction.h
@@ -50,7 +50,7 @@ enum ws_transaction_flags {
  */
 struct ws_transaction_command_list {
     size_t n; //!< @protected length of the command array
-    struct ws_command* cmds; //!< @protected Command objects
+    struct ws_statement* statements; //!< @protected Statements of the transaction
 };
 
 /**


src/objects/message/transaction.h
```diff
@@ -50,7 +50,7 @@ enum ws_transaction_flags {
  */
 struct ws_transaction_command_list {
     size_t n; //!< @protected length of the command array
-    struct ws_command* cmds; //!< @protected Command objects
+    struct ws_statements* statements; //!< @protected Statements of the transaction
```
matthiasbeyer: `ws_statements` Sounds wrong as typename here.

neithernut: Will fix _as soon as I'm done with my current task_.

neithernut: Interupted my work to fix this.

neithernut:
Info: I'm waiting for this, too. So Can I get an ACK already?

matthiasbeyer:
ACK.

PR #235, opened 2014-10-18
Add action manager processor

neithernut:
This PR introduces an initial implementation of the processor, which is the core of the aciton manager.
Targets #235.

### DIFF
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -101,8 +101,39 @@ ssize_t
 ws_processor_exec(
     struct ws_processor* self
 ) {
-    //!< @todo implement
-    return -1;
+    // keep a sentinel around for faster comparisons
+    struct ws_statement const* aend;
+    aend = self->commands->statements + self->commands->n;
+
+    while (self->pc < aend) {
+        // get the current pc and increment it afterwards
+        struct ws_statement const* cur = self->pc++;
+
+        // execute command
+        int res;
+        switch(cur->command->command_type) {
+        case regular:
+            res = exec_regular(self->stack, cur->command->func.regular,
+                               &cur->args);
+            break;
+
+        case special:
+            res = cur->command->func.special(self, &cur->args);
+            break;
+
+        default:
+            return -ENOTSUP;
+        }
+
+        if (res < 0) {
+            return res;
+        }
+    }
+
+    if (self->pc <= aend) {
+        return 0;
+    }
+    return self->pc - self->commands->statements;
 }
 
 size_t
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -52,6 +52,20 @@ ws_processor_prepare_args(
 __ws_nonnull__(1, 2)
 ;
 
+/**
+ * Execute a regaular command function
+ *
+ * This function executes a regular command function.
+ * This includes preparing and cleaning up the stack.
+ */
+static int
+exec_regular(
+    struct ws_processor_stack* stack,
+    ws_regular_command_func func,
+    struct ws_command_args const* const args
+)
+__ws_nonnull__(1, 2, 3)
+;
 
 /*
  *
@@ -148,5 +162,46 @@ ws_processor_prepare_args(
     }
 
     return 0;
+}
+
+static int
+exec_regular(
+    struct ws_processor_stack* stack,
+    ws_regular_command_func func,
+    struct ws_command_args const* const args
+) {
+    union ws_value_union* top = NULL;
+    if (args->num <= 0) {
+        return -EINVAL;
+    }
+
+    int res;
+
+    // prepare the stack for command execution
+    if (&args->vals) {
+        // the parameters are given explicitely
+        top = ws_processor_stack_top(stack);
+        res = ws_processor_prepare_args(stack, args);
+        if (res < 0) {
+            return res;
+        }
+    } else {
+        // the parameters are given implicitely
+        top = (union ws_value_union*)
+              ws_processor_stack_value_at(stack, -(args->num));
+    }
+
+    // we somehow ended up with a bad stack
+    if (!top) {
+        return -EINVAL;
+    }
+
+    res = func(top);
+    if (res < 0) {
+        return res;
+    }
+
+    // pop the values
+    return ws_processor_stack_pop(stack, args->num - 1);
 }
 
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -27,3 +27,36 @@
 
 #include "action/processor.h"
 
+int
+ws_processor_init(
+    struct ws_processor* self,
+    struct ws_processor_stack* stack,
+    struct ws_transaction_command_list* commands
+) {
+    //!< @todo implement
+    return -1;
+}
+
+void
+ws_processor_deinit(
+    struct ws_processor* self
+) {
+    //!< @todo implement
+}
+
+ssize_t
+ws_processor_exec(
+    struct ws_processor* self
+) {
+    //!< @todo implement
+    return -1;
+}
+
+size_t
+ws_processor_jump(
+    struct ws_processor* self,
+    size_t value
+) {
+    //!< @todo implement
+    return 0;
+}
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -141,7 +141,14 @@ ws_processor_jump(
     struct ws_processor* self,
     size_t value
 ) {
-    //!< @todo implement
+    //calculate the current position
+    size_t target_pos = (self->pc - self->commands->statements) + value;
+
+    // perform bound check
+    if (target_pos > self->commands->n) {
+        return target_pos;
+    }
+
     return 0;
 }
 
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -25,9 +25,39 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
+
 #include "action/processor.h"
 #include "action/processor_stack.h"
 #include "objects/message/transaction.h"
+#include "values/union.h"
+
+/*
+ *
+ * Forward declarations
+ *
+ */
+
+/**
+ * Prepare arguments for command execution
+ *
+ * This function prepares the stack for command execution by copying everything
+ * to the right position.
+ */
+static int
+ws_processor_prepare_args(
+    struct ws_processor_stack* stack,
+    struct ws_command_args const* const args
+)
+__ws_nonnull__(1, 2)
+;
+
+
+/*
+ *
+ * Interface implementation
+ *
+ */
 
 int
 ws_processor_init(
@@ -69,3 +99,54 @@ ws_processor_jump(
     //!< @todo implement
     return 0;
 }
+
+
+/*
+ *
+ * Internal implementation
+ *
+ */
+
+static int
+ws_processor_prepare_args(
+    struct ws_processor_stack* stack,
+    struct ws_command_args const* const args
+) {
+    union ws_value_union* top = ws_processor_stack_top(stack);
+    size_t argc = args->num;
+
+    //!< @todo optimize for if arguments already are in the right position
+    {
+        int res = ws_processor_stack_push(stack, argc);
+        if (res < 0) {
+            return res;
+        }
+    }
+
+    // iterate over all the arguments and assemble the stack
+    while (argc--) {
+        // get the argument
+        struct ws_argument* cur_arg = args->vals + argc;
+
+        // determine the source of the argument
+        struct ws_value* src = NULL;
+        switch (cur_arg->type) {
+        case direct:
+            src = cur_arg->arg.val;
+
+        case indirect:
+            src = ws_processor_stack_value_at(stack, cur_arg->arg.pos);
+        }
+
+        if (!src) {
+            return -EINVAL;
+        }
+        int res = ws_value_union_init_from_val(top + argc, src);
+        if (res < 0) {
+            return res;
+        }
+    }
+
+    return 0;
+}
+
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -25,13 +25,5 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
-#ifndef __WS_COMMAND_PROCESSOR_H__
-#define __WS_COMMAND_PROCESSOR_H__
+#include "action/processor.h"
 
-/**
- */
-struct ws_command_processor {
-    //!< @todo implement
-};
-
-#endif // __WS_COMMAND_PROCESSOR_H__
--- a/src/action/processor.c
+++ b/src/action/processor.c
@@ -26,6 +26,8 @@
  */
 
 #include "action/processor.h"
+#include "action/processor_stack.h"
+#include "objects/message/transaction.h"
 
 int
 ws_processor_init(
@@ -33,15 +35,22 @@ ws_processor_init(
     struct ws_processor_stack* stack,
     struct ws_transaction_command_list* commands
 ) {
-    //!< @todo implement
-    return -1;
+    // initialize all the fields
+    self->stack         = stack;
+    self->commands      = commands;
+    self->pc            = commands->statements;
+    self->stack_frame   = ws_processor_stack_start_frame(stack);
+
+    // nothing can go wrong here
+    return 0;
 }
 
 void
 ws_processor_deinit(
     struct ws_processor* self
 ) {
-    //!< @todo implement
+    // restore the old frame
+    (void) ws_processor_stack_restore_frame(self->stack, self->stack_frame);
 }
 
 ssize_t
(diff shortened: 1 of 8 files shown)


src/action/processor.c
```diff
@@ -0,0 +1,196 @@
...
+/*
+ *
+ * Forward declarations
+ *
+ */
+
+/**
+ * Prepare arguments for command execution
+ *
+ * This function prepares the stack for command ecxecution by copying everything
```
matthiasbeyer: Typo.

src/action/processor.c
```diff
@@ -0,0 +1,196 @@
...
+ * to the right position.
+ */
+static int
+ws_processor_prepare_args(
```
matthiasbeyer, line 48: As this is a file-local function, please remove the `ws_` prefix.

src/action/processor.c
```diff
@@ -0,0 +1,196 @@
...
+ *
+ */
+
+int
+ws_processor_init(
+    struct ws_processor* self,
+    struct ws_processor_stack* stack,
+    struct ws_transaction_command_list* commands
+) {
+    // initialize all the fields
```
matthiasbeyer, line 82: init

matthiasbeyer:
I didn't do a _that_ careful review, but the things look good. That's _NOT_ an ACK. I will do a review tomorrow without beeing tired then!

src/action/processor.c
xVin, line 82: lol :+1:

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+
+// forward declarations
+struct ws_value;
+union ws_value_union;
+
+/**
+ * A processor's stack
+ */
+struct ws_processor_stack {
+    union ws_value_union* data; //!< @private basepointer of the stack
```
matthiasbeyer, line 47: Why does this need to be a `ws_value_union`? You have the base type, `ws_value`, which you can use as ptr here!

src/values/union.c
```diff
@@ -0,0 +1,82 @@
...
+
+    case WS_VALUE_TYPE_NIL:
+        ws_value_nil_init(&dest->nil);
+        return 0;
+
+    case WS_VALUE_TYPE_BOOL:
+        ws_value_bool_init(&dest->bool_);
+        {
+            bool buf = ws_value_bool_plain(&dest->bool_);
+        return ws_value_bool_set(&dest->bool_, buf);
```
matthiasbeyer: Indentation.

src/values/union.c
```diff
@@ -0,0 +1,82 @@
...
+            return ws_value_int_set(&dest->int_, buf);
+        }
+
+    case WS_VALUE_TYPE_STRING:
+        ws_value_string_init(&dest->string);
+        {
+            struct ws_string* buf = ws_value_string_get_str(&dest->string);
+            ws_value_string_set_str(&dest->string, buf);
+            ws_object_unref(&buf->obj);
+        }
```
matthiasbeyer, line 74: Here is a return missing, right?

src/action/processor_stack.h
neithernut, line 47: I think I made mayself clear about this when I _introduced_ the `ws_value_union` type:
Nope, I'll not add malloc calls for each nil and int value we need!

neithernut:
Before this eventually gets merged, I'd like to bring in a proposal, which may result in some "extra" work within this PR:

The overal idea is that functions receive their argumetns as values being pushed on the stack. Currently, you may either give arguments directly or by a position on the stack as it was before the command involcation. This PR introduces a TODO, which suggests that this may be optimized in a way that allows calling functions without moving around values to construct the stack, however, it's not possible to do this without introducing some clusterfuck.

However, if we allow a statement to specify that it will take the topmost `n` values of the stack as arguments, things are suddenly very nice.

TL;DR: We should have statements which specify the number of arguments without shipping the actual argument list.

matthiasbeyer:
What kind of benefits would result from a change like that, when talking of user interaction with the API?

neithernut:
Evaluation of expressions is a whole lot faster, and more intuitive for people who know Reverse Polish notation. It also eases things considerably if we come up with a DSL.

matthiasbeyer:
I assume that the request/transaction steps would look like this:
- I want 3 arguments from the stack for the next command
- The next command

right? If yes, this sounds sane to me. But it would also cause changes in the JSON parser, wouldn't it? If yes, I would recommend add such functionality later. (Execution/Parsing) Speed is not our only concern at this moment!

neithernut:
It's also a question of a sane interface!

I submitted a PR (#238). As you'll see, the changes in the interface are purely in semantics and can therefore be introduced without any problems emerging (as long as I perform the neccessary changes _first_, which I intent).

neithernut:
I rebased and added the new feature. in `ws_processor_exec()`

src/action/processor.c
```diff
@@ -0,0 +1,211 @@
...
+        struct ws_value* src = NULL;
+        switch (cur_arg->type) {
+        case direct:
+            src = cur_arg->arg.val;
+
+        case indirect:
+            src = ws_processor_stack_value_at(stack, cur_arg->arg.pos);
+        }
+
+        if (!src) return -EINVAL;
```
matthiasbeyer: Missing block around return statement here.

src/action/processor.c
```diff
@@ -0,0 +1,211 @@
...
+size_t
+ws_processor_jump(
+    struct ws_processor* self,
+    size_t value
+) {
+    //calculate the current position
+    size_t target_pos = (self->pc - self->commands->statements) + value;
+
+    // perform bound check
+    if (target_pos >self->commands->n) {
```
matthiasbeyer: Missing space after `>`

src/action/processor.h
```diff
@@ -0,0 +1,104 @@
...
+;
+
+/**
+ * Perform a jump (by incrementing the program counter)
+ *
+ * @return 0 if the jump was successful, a positive value if the jump would
+ *           result in a position outside the current statement list.
+ */
+size_t
+ws_processor_jump(
```
matthiasbeyer, line 103: Why needs this to be exported?

src/action/processor.h
```diff
@@ -0,0 +1,104 @@
...
+ws_processor_exec(
+    struct ws_processor* self //!< command processor context
+)
+__ws_nonnull__(1)
+;
+
+/**
+ * Perform a jump (by incrementing the program counter)
+ *
+ * @return 0 if the jump was successful, a positive value if the jump would
```
matthiasbeyer, line 99: `"would"` ? So it doesn't actually do something if the jump would result out of the statement list? Please clearify!

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+    size_t size; //!< @private size of the stack
+    size_t top; //!< @private top of the stack, as position from the basepointer
+    size_t frame; //!< @private base pointer of the current frame
+};
```
matthiasbeyer, line 51: As all fields in the processor stack type: Why not hiding this type implementation?

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+    size_t slots //!< number of slots to pop
+)
+__ws_nonnull__(1)
+;
+
+/**
+ * Get the top of the stack
+ *
+ * This function returns the top of the stack, which actually is an invalid
+ * value.
```
matthiasbeyer, line 112: Why would you want an invalid value from the stack? Maybe add some more information here why it can be useful.

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+ * If you subtract from that pointer, you get valid values _on_ the stack.
+ * At some point, you reach the bottom of the stack.
```
matthiasbeyer, line 115: How to identify the bottom of the stack when doing this?

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+    struct ws_processor_stack* self //!< the stack
+)
+__ws_nonnull__(1)
+;
+
+/**
+ * Get the bottom of the stack
+ *
+ * This function returns the bottom of the stack.
+ * If you add to that pointer, you move toward the top of the stack.
```
matthiasbeyer, line 130: How to identify that we are on the last valid value?

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Get a value at a given position
+ *
+ * @return value at given position
+ */
+struct ws_value*
+ws_processor_stack_value_at(
+    struct ws_processor_stack* self, //!< the stack
+    ssize_t pos //!< position of value
```
matthiasbeyer, line 149: Why can this be negative? Please add a short note! (I understand why, but well...)

src/action/processor_stack.h
```diff
@@ -0,0 +1,176 @@
...
+ws_processor_stack_start_frame(
+    struct ws_processor_stack* self //!< the stack
+)
+__ws_nonnull__(1)
+;
+
+/**
+ * Restore a frame
+ *
+ * Call this function to restore a frame previously started using
```
matthiasbeyer, line 169: What does "restore" mean in this context?

src/action/processor_stack.c
```diff
@@ -0,0 +1,230 @@
...
+#include <malloc.h>
+#include <string.h>
+
+#include "action/processor_stack.h"
+#include "values/nil.h"
+#include "values/union.h"
+#include "values/value.h"
+
+
+#define INITIAL_STACK_SIZE (4)
```
matthiasbeyer, line 39: Is this actually a sane value? I consider it as much to less, but well, we can optimize that later on very easily, I guess.

matthiasbeyer, line 39: Ah, just saw you do a `*2` reallocation on resize. Nuff said.

src/action/processor.h
neithernut, line 103: You use this function in the "jump" command.

src/action/processor_stack.c
```diff
@@ -0,0 +1,230 @@
...
+}
+
+void
+ws_processor_stack_pop(
+    struct ws_processor_stack* self,
+    size_t slots
+) {
+    // check whether there are at least that many values on the stack
+    if (slots > (self->top - self->frame)) {
+        slots = self->top - self->frame;
```
matthiasbeyer: If I understand things right: If `slots` is bigger than the number of values on the stack, we only pop the maximum value. This should be stated in the documentation of the function, I guess.

src/action/processor_stack.c
```diff
@@ -0,0 +1,230 @@
...
+    size_t new_top  = self->top - slots;
+    while ((new_top > new_size) && (new_top > INITIAL_STACK_SIZE)) {
+        new_size /= 2;
+    }
+
+    // deinitialize all the values on the way down
+    size_t cur = self->top;
+    while (cur > new_top) {
+        --cur;
+        ws_value_deinit(&self->data[cur].value);
```
matthiasbeyer, line 132: You deinitialize here, but shouldn't be the values be initialized to `nil` afterwards? That's what you do when growing the stack, so I assume you should do it here, too, right?

src/action/processor_stack.c
```diff
@@ -0,0 +1,230 @@
...
+    self->top = new_top;
+
+    // check whether we may resize
+    if (new_size != self->size) {
+        // try to reallocate
+        union ws_value_union* new_data;
+        new_data = realloc(self->data, sizeof(*(self->data)) * new_size);
+        if (!new_data) {
+            // noncritical error
+            return;
```
matthiasbeyer: Would be shorter if the if wouldn't use an early return and you would set the members if the reallocation succeeded. _duckandrun_

src/action/processor_stack.c
```diff
@@ -0,0 +1,230 @@
...
+
+    return self->data;
+}
+
+struct ws_value*
+ws_processor_stack_value_at(
+    struct ws_processor_stack* self,
+    ssize_t pos
+) {
+    size_t top = self->top;
```
matthiasbeyer: I would embedd the `self->top` into the code below, it's not that much overhead to read.

src/values/union.h
```diff
@@ -25,6 +25,16 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+
+/**
+ * @addtogroup values "Value types"
+ *
+ * @{
+ */
+
```
matthiasbeyer, line 34: Maybe we should add another documentation group within the values group for internal stuff. (I consider that the union value type is not exported through the API, right?

src/values/union.h
```diff
@@ -37,8 +47,21 @@
...
-    struct ws_value_bool    bool;   //!< bool value
-    struct ws_value_int     int;    //!< int value
+    struct ws_value_bool    bool_;  //!< bool value
+    struct ws_value_int     int_;   //!< int value
     struct ws_value_string  string; //!< string value
 };
 
+/**
+ * Initializes the union by copying another value
+ */
```
matthiasbeyer, line 61: Missing: Return documentation, documentation why this does nothing if the value type is either "value" or a "non-value" type.

matthiasbeyer:
Finally I managed to do a complete review on this. The code looks really sane for me and I like what you introduced. So please, don't feel bad about my comments here!

src/action/processor_stack.c
neithernut: No, it wouldn't.

matthiasbeyer: It would result in 3 less lines in the function.

src/action/processor_stack.h
neithernut, line 51: Because it will be allocated on the stack, not on the heap.

src/action/processor_stack.h
neithernut, line 169: Seriously? Should be pretty obvious.

src/action/processor_stack.c
neithernut, line 132: Nope, values which are not on the stack _must_ be invalid, so we can determine the number of arguments a function gets by reading up to the first invalid argument.

src/values/union.h
neithernut, line 34: Of course it's not. But managing the documentation groups is nothing that should go in this PR.

src/action/processor_stack.h
neithernut, line 130: `WS_VALUE_TYPE_NONE`

src/action/processor_stack.h
neithernut, line 149: **not on this scope**
This belongs in an more extensive external documentation.

src/action/processor_stack.h
neithernut, line 115: Coparision to the pointer you get from `ws_processor_stack_bottom()`

src/action/processor_stack.h
matthiasbeyer, line 149: So add an `@see` to this documentation.

src/action/processor_stack.h
matthiasbeyer, line 130: Should be in the doc.

src/action/processor_stack.h
neithernut, line 149: I have to write that documentation **first**.

neithernut:
I added the first chunk of documentation of stack semantics. More to come.

src/action/processor_stack.h
matthiasbeyer, line 149: Of course.

src/action/stack.md
```diff
@@ -0,0 +1,82 @@
...
+ |        | <- top
+ | FOO    |
+ | BAR    | <- argument of the function implementation
+ | BAZ    |
+ | DUCK   | <- bottom
+ +--------+
+
+After a function invocation, all the arguments are popped from the stack except
+the first one, which is the return value of the function.
+
```
matthiasbeyer, line 84: This belongs into `doc/` in my opinion. I will open an Issue for this to discuss.

neithernut:
I outsources a few lines of code from `ws_processor_exec()` to another function and fixed and expanded the documentation a little.
Please review again and tell me if there are any questions left.

neithernut:
Updated because of #243.

src/action/stack.md
```diff
@@ -0,0 +1,114 @@
...
+top, `-1` being the topmost value, `-2` the next lower one and so on.
+The following graph shows a stack with four items and the position of each one
+of them in both direct and relative addressing:
+
+    +------+
+    |      |         <- top
+    | FOO  | -1 or 3
+    | BAR  | -2 or 2
+    | BAZ  | -3 or 1
+    | DUCK | -4 or 0 <- bottom
```
matthiasbeyer, line 22: :+1: for "duck"

src/action/processor_stack.h
matthiasbeyer, line 169: It is now, after I read your documentation. Please add a `@see` in the function documentation, refering to the documentation document.

neithernut, line 169: The type documentation already has a `@see` tag.

matthiasbeyer:
ACK.

For the documentation thing: I will add a note in the related issue.

PR #237, opened 2014-10-19
Add value name table

matthiasbeyer:
This PR adds a value-type-name table and a function to get a value
type identifier from a string.

It also adds a arithmetical macro as helper.

### DIFF
--- a/src/util/arithmetical.h
+++ b/src/util/arithmetical.h
@@ -62,6 +62,13 @@
  */
 #define CLAMP(x,y,z) MAX(x,MIN(y,z))
 
+/**
+ * Get the length of an array
+ *
+ * @returns The length of the array
+ */
+#define ARYLEN(x) (sizeof((x))/sizeof((x)[0]))
+
 #endif // __WS_UTIL_ARITHMETICAL_H__
 
 /**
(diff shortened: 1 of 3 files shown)


src/values/value_type.c
```diff
@@ -0,0 +1,58 @@
...
+
+enum ws_value_type
+ws_value_type_from_value_name(
+    char const* name
+) {
+    for (enum ws_value_type i = 0; i < LEN(WS_VALUE_TYPE_NAMES); i++) {
+        if (0 == strcmp(WS_VALUE_TYPE_NAMES[i], name)) {
+            return i;
+        }
+    }
```
neithernut, line 54: Would prefer a downcounting while loop, but this is just a matter of taste.

src/util/arithmetical.h
```diff
@@ -62,6 +62,13 @@
  */
 #define CLAMP(x,y,z) MAX(x,MIN(y,z))
 
+/**
+ * Get the length of an array
+ *
+ * @returns The length of the array
+ */
+#define LEN(x) (sizeof((x))/sizeof((x)[0]))
```
neithernut: `LEN` is ambiguous, I recommend renaming it to something like `aray_len`

neithernut:
ACK.

PR #242, opened 2014-10-20
Fix union header

neithernut:
This is a fix which was previously included in the PR #235.


neithernut:
Can I get an ACK?

src/values/union.h
```diff
@@ -37,8 +47,10 @@
...
     struct ws_value_nil     nil;    //!< nil value
-    struct ws_value_bool    bool;   //!< bool value
-    struct ws_value_int     int;    //!< int value
+    struct ws_value_bool    bool_;  //!< bool value
+    struct ws_value_int     int_;   //!< int value
     struct ws_value_string  string; //!< string value
 };
 
+#endif //__WS_VALUES_UNION_H__
+
```
matthiasbeyer, line 56: You must close the documentation group here.

``` c
/**
 * @}
 */
```

neithernut, line 56: Good to know, I'll close this PR anyway because other things are missing.
I'll open another one.

neithernut:
Close unmerged as the scope of the PR is offset from what I really want/need.

matthiasbeyer:
Also: Why is this tagged with the bug label?

matthiasbeyer:
huh,... okay

PR #243, opened 2014-10-20
Add union features

neithernut:
This PR introduces fixes to the `src/values/union.h` header and adds a function. Both the PR #235 and my current assignment need this to function. (it's acutally ripped off of #235).
This is kind of urgent, so please give me an ACK as soon as possible.


matthiasbeyer:
ACK. Looks really good.

Issue #71, opened 2014-09-30
Add value type "object_id"


neithernut:
We need `ws_value_object_id`.

neithernut:
Blocks #64.

matthiasbeyer:
Q: has this to hold a `ws_object` or a `ws_wayland_obj` ?

neithernut:
It has an object's reference, but it deserializes to an id.

matthiasbeyer:
You mean a hash of a object to identify it in a set?

neithernut:
Nope, not a hash but some sort of id to find it in the world set/object model (which implies that it's only relevant for object types which are in the object model). Should probably be the wayland id.

PR #252, opened 2014-10-20
Bootstrapped ws_value_object_id type

manuelmessner:
Targets #71

### DIFF
--- a/src/values/object_id.c
+++ b/src/values/object_id.c
@@ -40,6 +40,13 @@ void
 ws_value_object_id_init(
     struct ws_value_object_id* self
 ) {
+    if (!self) {
+        return;
+    }
+
+    ws_value_init(&self->val);
+    self->val.type = WS_VALUE_TYPE_OBJECT_ID;
+    self->val.deinit_callback = value_object_id_deinit;
 }
 
 struct ws_object*
--- a/src/values/object_id.c
+++ b/src/values/object_id.c
@@ -65,5 +65,19 @@ ws_value_object_id_set(
     struct ws_value_object_id* self,
     struct ws_object* obj
 ) {
+    if (!self || !obj) {
+        return;
+    }
+
+    if (self->obj) {
+        ws_object_unref(self->obj);
+    }
+
+    struct ws_object* tmp = getref(obj);
+    if (!tmp) {
+        return;
+    }
+
+    self->obj = tmp;
 }
 
--- a/src/values/object_id.c
+++ b/src/values/object_id.c
@@ -53,6 +53,11 @@ struct ws_object*
 ws_value_object_id_get(
     struct ws_value_object_id* self
 ) {
+    if (!self || !self->obj) {
+        return NULL;
+    }
+
+    return getref(self->obj);
 }
 
 void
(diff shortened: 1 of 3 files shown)


manuelmessner:
@neithernut 
Did you mean something like this?

What exactly do you mean by `id`?

src/values/object_id.h
```diff
@@ -35,6 +35,64 @@
...
+/**
+ * Allocate a new ws_value_object_id object and initialize it
+ *
+ * @memberof ws_value_object_id
+ *
+ * @return a new, allocated and initialized ws_value_object_id object or NULL on
+ * failure
+ */
+struct ws_value_object_id*
+ws_value_object_id_new(void);
```
neithernut: We don't need the allocation function for values, you can ommit this.

neithernut:
Yep, pretty much what we need. Just had this one annotation.

manuelmessner:
Ok i'll fix that

manuelmessner:
@neithernut 
Can I start to implement?

src/values/object_id.h
```diff
@@ -35,6 +35,52 @@
...
+/**
+ * Set the `obj` member of a given ws_value_object_id object to a reference to
+ * a given ws_object object
+ *
+ * @memberof ws_value_object_id
+ */
+void
+ws_value_object_id_set(
+    struct ws_value_object_id* self,
+    struct ws_object* str
```
matthiasbeyer: Why is this called `str` ?

neithernut:
Yes, you can start to implement the interface.

src/values/object_id.h
manuelmessner: Because I was thinking for a moment what else we could need and thought about strings as example. so I accidently wrote str.
It's fixed.

matthiasbeyer: I don't see the fix here, but I assume you push this later... so everything right!

manuelmessner: Yep did'nt pushed yet

manuelmessner:
@neithernut @matthiasbeyer 
Review pls

matthiasbeyer:
Will review tomorrow. Good night, ladies and gentlemen!

manuelmessner:
Good night :)

src/values/object_id.c
```diff
@@ -25,5 +25,68 @@
...
+
+struct ws_object*
+ws_value_object_id_get(
+    struct ws_value_object_id* self
+) {
+    if (!self || !self->obj) {
+        return NULL;
+    }
+
+    return self->obj;
```
neithernut: If you bother to check whether `self->obj` is `NULL`, it would also make sense to return it `getref()`'d.

neithernut:
The work makes a good impression so far. I think this can be merged rrather soon.

src/values/object_id.c
manuelmessner: Ok, I'll add that.

manuelmessner:
Ok I clean the history tomorrow mornig

neithernut:
ACK.

matthiasbeyer:
ACK.

PR #253, opened 2014-10-20
Add ws_value instance from name

matthiasbeyer:
This is a convenience function which is good for the
implementation of the JSON parser, as it reduces overhead in
other PRs.


neithernut:
Seems legit.

matthiasbeyer:
Feel free to merge then.

PR #261, opened 2014-10-21
Added wl_output global to ws_monitor

TheNeikos:


src/compositor/monitor.c
```diff
@@ -120,6 +135,68 @@ ws_monitor_surfaces(
...
+    struct wl_client* client,
+    void* data,
+    uint32_t version,
+    uint32_t id
+) {
+
+    struct ws_monitor* monitor = data;
+
+    if (version >= 2) {
+        version = 2;
```
matthiasbeyer, line 152: Why are you doing this? Please add documentation!

matthiasbeyer:
ACK besides the one comment.

src/compositor/monitor.c
```diff
@@ -64,6 +67,19 @@ ws_monitor_cmp(
...
+    uint32_t version,
+    uint32_t id
+);
+
+static int
+publish_modes(
+    void* _data,
+    void const* _mode
+);
 /*
```
neithernut, line 84: Maybe add a space here?

src/compositor/monitor.c
```diff
@@ -120,6 +136,74 @@ ws_monitor_surfaces(
...
+    struct wl_client* client,
+    void* data,
+    uint32_t version,
+    uint32_t id
+) {
+
+    struct ws_monitor* monitor = data;
+
+    if (version >= 2) {
+        // We only support from 2 and ongoing, so let's say that!
```
neithernut, line 151: You meant to say: "2 or lower".

TheNeikos, line 151: Nope! Just that version 3 for example would add a new function call (since they don't break the api contract of previous versions). We only implement version 2 though.

neithernut, line 151: Ahhh, yes.

neithernut:
ACK.

matthiasbeyer:
Is this code ready?

TheNeikos:
I can't assert that. There is no way I know of to test it. Like with everything in the compositor we will only know it doesn't work once it bugs out. (Thanks to the awesome documentation)

PR #266, opened 2014-10-21
Fixed segfault in ws_monitor_publish

TheNeikos:
This was a misunderstanding (due to no documentation) and it seems that
monitor->resource will only be created once a client requests it, this commit
removes it from the compositor module.

### DIFF
--- a/src/compositor/module.c
+++ b/src/compositor/module.c
@@ -459,7 +459,6 @@ populate_connectors(void) {
 
 insert:
         new_monitor->id = i;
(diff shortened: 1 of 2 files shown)


neithernut:
I looked at the issue in _some_ detail and I could verify both the problem and the solution. There is a bit more to this, though. I think I'll discuss this with you tomorrow --if I have a clear head.
ACK for now (will be merged, but another PR will hopefully follow).

PR #273, opened 2014-10-22
Add preliminary evdev discovery

TheNeikos:


matthiasbeyer:
First a basic question: Why is this in a new module? Isn't it a compositor thing?

TheNeikos:
Input is not something the compositor should handle. It should just compose (render the final image).

matthiasbeyer:
Hm. @neithernut what are you thinking? The problem is that it differs from our inital planning of the modules, you know? That's why I ask.

neithernut:
We already decided to have an input module, if you remember...

src/input/module.c
```diff
@@ -0,0 +1,180 @@
...
+        return -1;
+    }
+
+
+    while ((dir = readdir(d)) != NULL) {
+        if (!filter_name(dir->d_name)) {
+            continue;
+        }
+        char full_path[100] = "/dev/input/";
+        strcat(full_path, dir->d_name);
```
neithernut: would ust `strncat` here.

matthiasbeyer:
Huh, right... but that was a long time ago, right? So, well... I'll review now.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+#include "objects/set.h"
+
+static struct ws_logger_context log_ctx = { .prefix = "[Input] " };
+
+static struct ws_set devices;
+
+#define INPUT_PATH "/dev/input/"
+#define DEVICE_NAME "event"
+
+static void
```
matthiasbeyer, line 54: Overall function documentation missing.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+        event = (struct inotify_event*)
+            (((char*) event) + event->len + sizeof(*event));
+    }
+
+
+}
+
+static bool
+filter_name(
+    const char* name
```
matthiasbeyer, line 94: Overall function documentation missing as well as for this parameter.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+) {
+    if (!(strcmp(".", name) && strcmp("..", name))) {
+        return false;
+    }
+    if (strncmp(DEVICE_NAME, name, sizeof(DEVICE_NAME))) {
```
matthiasbeyer: Do you really need an `strncmp()` call here? `strcmp()` should work fine, as `DEVICE_NAME` is null-terminated, isn't it?

matthiasbeyer: Also you can do a `return (0 == strcmp(...` here.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+        return false;
+    }
+    return true;
+}
+
+static int
```
matthiasbeyer, line 112: Function documentation missing.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+find_initial_devices()
```
matthiasbeyer: `void` inside the parentheses missing.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+{
```
matthiasbeyer: You can put the open bracket on the line above if you do not get any arguments here:

``` C
void
foo(void) {
```

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+    DIR* d;
+    struct dirent* dir;
+
+    d = opendir(INPUT_PATH);
+
+    if (!d) {
+        ws_log(&log_ctx, LOG_CRIT, "Could not open input device director");
```
matthiasbeyer: "director"? You mean "directory" here, right?

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+        return -1;
+    }
+
+
+    while ((dir = readdir(d)) != NULL) {
+        if (!filter_name(dir->d_name)) {
+            continue;
+        }
+        char full_path[100] = INPUT_PATH;
```
matthiasbeyer: Where comes this magic constant from? Please document it.

src/input/module.c
```diff
@@ -0,0 +1,183 @@
...
+    if (watchfd < 0) {
+        ws_log(&log_ctx, LOG_CRIT, "Could not watch input devices");
+        return watchfd;
+    }
+
+    struct ev_io* watcher = calloc(1, sizeof(*watcher));
+    ev_io_init(watcher, watch_devices, inotfd, EV_READ);
+
+    struct ev_loop* default_loop = ev_default_loop(EVFLAG_AUTO);
+    if (!default_loop) return -1;
```
matthiasbeyer: Brackets missing.

src/input/module.c
TheNeikos: Nope, if S1 < S2 it will return a wrong value

neithernut:
ACK.

matthiasbeyer:
ACK.

PR #274, opened 2014-10-22
Logical commands

matthiasbeyer:
This PR adds the logical commands. By now, only the stubs are present.
Is there anything missing here? Please tell me.

@waysome/hackers

### DIFF
--- a/src/command/logical.c
+++ b/src/command/logical.c
@@ -49,8 +49,24 @@ ws_builtin_cmd_land(
     union ws_value_union* args
 
 ) {
-    //!< @todo implement
-    return -1;
+    struct ws_value* base = &args->value;
+
+    while (ws_value_get_type(&args->value) != WS_VALUE_TYPE_NONE) {
+        if (ws_value_is_true(&args->value)) {
+            args++;
+            continue;
+        }
+
+        ws_value_deinit((struct ws_value*) base);
+        ws_value_bool_init((struct ws_value_bool*) base);
+        ws_value_bool_set((struct ws_value_bool*) base, false);
+        return 0;
+    }
+
+    ws_value_deinit((struct ws_value*) base);
+    ws_value_bool_init((struct ws_value_bool*) base);
+    ws_value_bool_set((struct ws_value_bool*) base, true);
+    return 0;
 }
 
 int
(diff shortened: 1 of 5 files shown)


neithernut:
Sufficient for now.

matthiasbeyer:
I'm ready for review here.

neithernut:
You forgot to deinit the target value in case it's not an bool.
Also: I think it would be better if the commands were shorter in name, e.g. `lnot`, `land`, `lnand`

matthiasbeyer:
Hm. I don't see any benefits from shorter names and I don't see any drawbacks by the longer ones.

neithernut:
Users have to **type** them.

matthiasbeyer:
Ah, did not thought about that they get exported directly. Okay, ... But I guess users will _not_ have to type them as there will be libraries. I don't think anyone will write a transaction by hand.

But kay, I will fix this.

src/command/logical.c
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        ws_value_deinit((struct ws_value*) base);
+        ws_value_bool_init((struct ws_value_bool*) base);
+        ws_value_bool_set((struct ws_value_bool*) base, false);
+        return 0;
+    }
+
+    return 0;
+}
```
neithernut, line 70: This function fails if all values are falsy.

src/command/logical.c
```diff
@@ -0,0 +1,146 @@
...
+        if (ws_value_is_true(&args->value)) {
+            ws_value_deinit((struct ws_value*) base);
+            ws_value_bool_init((struct ws_value_bool*) base);
+            ws_value_bool_set(&base->bool_, true);
+            return 0;
+        }
+        args++;
+    }
+
+    ws_value_bool_set(&base->bool_, false);
```
neithernut, line 106: You did forget the re-initialisation here.

src/command/logical.c
```diff
@@ -0,0 +1,146 @@
...
+                return 0;
+            }
+
+            have_true = true;
+        }
+
+        args++;
+    }
+
+    ws_value_bool_set(&args->bool_, have_true);
```
neithernut: You did forget the reinitialization here.

neithernut:
ACK.

PR #276, opened 2014-10-22
Return -errno insead of return value in connector.c

neithernut:
This little PR fixes some issues arising with using negative error numbers returned by functions by making the functions return an error number we can actually use to do things.


xVin:
Ack.

matthiasbeyer:
ACK.

Issue #282, opened 2014-10-22
Add function for verifying type of object


neithernut:
We need something like `is_instance_of`.

PR #283, opened 2014-10-22
Add query for whether an object is an instance of a type

neithernut:
Targets #282.

### DIFF
--- a/src/objects/object.c
+++ b/src/objects/object.c
@@ -596,6 +596,27 @@ iter_next:
     return self->uuid;
 }
 
+bool
+ws_object_is_instance_of(
+    struct ws_object const* self,
+    ws_object_type_id* type
+) {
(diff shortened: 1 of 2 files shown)


src/objects/object.c
```diff
@@ -596,6 +596,27 @@ ws_object_uuid(
...
+bool
+ws_object_is_instance_of(
+    struct ws_object const* self,
+    ws_object_type_id* type
+) {
+    ws_object_type_id* curtype = self->id;
+
+    // iterate over all the base types
+    while (curtype) {
+        if (curtype == type) {
```
matthiasbeyer, line 608: I guess you missed that `ws_object` does inherit from itself, and you fogot to check it here. So this would (as it currently is) be an endless loop.

matthiasbeyer:
ACK.

Issue #272, opened 2014-10-22
Remove deprecated methods


matthiasbeyer:
We currently have the methods
- `ws_object_unlock_read()`
- `ws_object_unlock_write()`

These should be removed from the source code as well as all calls to them (and be replaced with `ws_object_unlock()` calls).

PR #291, opened 2014-10-23
Remove ws object unlock read write

manuelmessner:
Targets #272


manuelmessner:
@matthiasbeyer 
Can you review pls?

There was just the implementation no call.

neithernut:
Yep, since we already migrated away from this (with `HARD_MODE=ON`, it refuses to build anyway if you use on of these --at least with `clang`).
ACK.

PR #260, opened 2014-10-21
Add JSON deserializer yajl callbacks

matthiasbeyer:
This PR adds the JSON deserializer yajl callbacks with implementation.

Please review and tell me what is missing and what can be
optimized here.

PR #292, opened 2014-10-23
Allow yajl to use C runtime functions (malloc, realloc, free)

matthiasbeyer:
It turned out that yajl uses malloc()/free()/realloc() if we do
not set these callbacks. So we do this here.

This is an rip from #260.


neithernut:
ACK.

PR #293, opened 2014-10-23
Fix some json deserializer state helpers

matthiasbeyer:
Another rip from #260 

This removes the outdated helpers, as functionality was brought up
in several PRs which makes them obsolete.

It also fixes some minor things on the internal datastructure.


neithernut:
ACK.

PR #295, opened 2014-10-23
Extend statement initializer by the command name

neithernut:
I realized that it makes sense to initialize the command using a name. It's ok for the WIP JSON deserializer, since the parser wil see the command name before it sees the arguments _always_. Other formats should take care to have the command name before the argument list, which should be the case anyway.

This has two concrete benefits:
- The parsers don't need to know anything about `command/command.h`.
- Statement construction get's even easier.

### DIFF
--- a/src/command/statement.c
+++ b/src/command/statement.c
@@ -27,6 +27,7 @@
 
 #include <errno.h>
 
+#include "command/command.h"
 #include "command/statement.h"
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Please fix typo in both commit message and PR title.

matthiasbeyer:
Everything else: ACK.

PR #297, opened 2014-10-24
Fix command argument appending function

neithernut:
Previously we always returned the first element. Now we return the
last one (the one we just allocated), which was intented in the first
place.


TheNeikos:
I don't see anything wrong, ACK

matthiasbeyer:
Me too.  ACK.

PR #299, opened 2014-10-25
Fixes/Preparation for json deserializer

matthiasbeyer:
This is another rip from #260 which fixes a bunch of things. #260 itself needs
to be rebased on this and could then be merged, so please review carefully.

### DIFF
--- a/src/serialize/json/deserializer_state.c
+++ b/src/serialize/json/deserializer_state.c
@@ -55,8 +55,12 @@ deserialize_state_new(
     }
 
     state->current_state    = STATE_INIT;
-    state->uid              = 0;
     state->tmp_statement    = NULL;
+
+    state->nboxbrackets     = 0;
+    state->ncurvedbrackets  = 0;
+
+    state->id               = 0;
 
     return state;
 }
--- a/src/objects/message/transaction.c
+++ b/src/objects/message/transaction.c
@@ -85,7 +85,7 @@ ws_transaction_new(
         return NULL;
     }
 
-    if (!ws_message_init(&t->m, id)) {
+    if (0 != ws_message_init(&t->m, id)) {
         free(t);
         return NULL;
     }
--- a/src/objects/message/transaction.c
+++ b/src/objects/message/transaction.c
@@ -155,26 +155,28 @@ ws_transaction_push_statement(
         }
 
         t->cmds->statements = NULL;
-        t->cmds->n          = 0;
-    } else {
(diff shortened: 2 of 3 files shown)


src/command/statement.c
```diff
@@ -163,6 +163,8 @@ command_args_append(
     }
 
     self->vals = newargs;
-    return newargs + self->num++;
+    struct ws_argument* ret = &newargs[self->num];
+    self->num++;
+    return ret;
```
neithernut: Please describe **exactely** what problems you did have with the previous solution, because from a functional view, there is **no** change introduced by this commit --at least no change which **I** would be aware of.

If you fail to justify this changeset, I'll refuse to ACK this commit.
If you manage to justify this changeset, there is one termination point in the same function which may also be malicious.

---

Off topic: obviously, our sense in aestetics differs a great deal here.

matthiasbeyer: I just re-run the tests without this patch and they still succeed, so it was another patch which fixed my problems. I will remove this patch from the changeset now.

src/serialize/json/deserializer_state.c
```diff
@@ -55,9 +55,13 @@ deserialize_state_new(
     }
 
     state->current_state    = STATE_INIT;
-    state->uid              = 0;
     state->tmp_statement    = NULL;
 
+    state->nboxbrackets     = 0;
+    state->ncurvedbrackets  = 0;
```
neithernut, line 61: You _could_ use one integer for both maps and arrays...
But this would be a rather aestetical (and thus maybe unnecessary) change.

neithernut:
Woops, mis-clicked.

src/command/command.h
```diff
@@ -60,7 +60,7 @@ typedef int (*ws_special_command_func)(struct ws_processor*,
  * This datatype represents a command, which may be invoked in a transaction.
  */
 struct ws_command {
-    char const* const name; //!< @public name of the command
+    char const* name; //!< @public name of the command
```
neithernut: I think this changeset is a result of bad usage.

You can omit this changeset now, since you don't even have anything to do with a command anymore (you sould use `ws_statement_init()` anyways). You shouldn't even include the `command.h` header.

matthiasbeyer:
Fixed what you said (btw I'm keeping #260 up to date and I remove the patches from there when removing them here. When rebasing, the most stuff will be thrown out automatically, but not the patches I removed here, so I'm doing this by hand every time, so I don't have to do this afterwards).

neithernut:
ACK.

Issue #288, opened 2014-10-22
Add binary commands


matthiasbeyer:
In relation to #268 , one should add the binary operations/commands.

I will do this this weekend.

matthiasbeyer:
@neithernut Please add the appropriate Milestone here.

PR #301, opened 2014-10-25
Add implementation for binary command builtins

matthiasbeyer:
Targets #288

### DIFF
--- a/src/command/binary.c
+++ b/src/command/binary.c
@@ -132,8 +132,11 @@ ws_builtin_cmd_bnor(
     union ws_value_union* args
 
 ) {
-    //!< @todo implement
-    return 0;
+    int ret = ws_builtin_cmd_bor(args);
+    if (ret != 0) {
+        return ret;
+    }
+    return bnot(args);
 }
 
 int
--- a/src/command/binary.c
+++ b/src/command/binary.c
@@ -28,6 +28,7 @@
 #include <errno.h>
 
 #include "command/binary.h"
+#include "command/util.h"
 #include "values/value_type.h"
 #include "values/union.h"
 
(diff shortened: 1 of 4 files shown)


neithernut:
Binary operations should only operate on integer types (any fail for _anything_ else). For boolean values, the logical commands should be used.

matthiasbeyer:
Fixed, please review again. Will be able to do squashing at 11PM CET or something.

src/command/binary.c
```diff
@@ -0,0 +1,187 @@
...
+    union ws_value_union* args
+) {
+    if (ws_value_get_type(&args->value) == WS_VALUE_TYPE_NONE ||
+            ws_value_get_type(&args[1].value) != WS_VALUE_TYPE_NONE) {
+        // There is no argument or there are more arguments than one
+        return -EINVAL;
+    }
+
+    return bnot(args);
+}
```
neithernut, line 72: Mh... this function does _exactly_ what `bnot()` does (the checks are unneccessary btw.), so why don't you just integrate `bnot()`'s body into this function and use this one instead from the other ones?

src/command/binary.c
```diff
@@ -0,0 +1,187 @@
...
+            return -EINVAL;
+        }
+
+        intmax_t buf = ws_value_int_get(&args->int_);
+        buf &= ws_value_int_get(&base->int_);
+        ws_value_int_set(&base->int_, buf);
+    }
+
+    return 0;
+}
```
neithernut, line 94: There are residuals from when you also considered boolean values. (the entire `type` stuff. It's not needed any more) Same holds true for all the other functions.

src/command/binary.c
matthiasbeyer, line 72: Why are the checks unneccessary? If that's really the case, I am willing to do this, of course.

neithernut, line 72: Because of the checks you perform in `bnot()`.

matthiasbeyer, line 72: Not exactly: This function also checks if there was really just one argument passed here. The `bnot()` function does not (I did this to be able to use to use the `bnot()` function also for the other `not-...` commands).

neithernut:
You could migrat to the new utility macros now.

matthiasbeyer:
Done.

neithernut:
Looks fine now.

matthiasbeyer:
I'm rebasing now.

matthiasbeyer:
Done.

neithernut:
ACK.

Issue #298, opened 2014-10-24
Remove init callback


matthiasbeyer:
The "init" callback for the object type (and all subtypes) does not make sense at all. We should get rid of it until some point.

I assigned this to @xVin because it is not that urgent (no critic here)!

xVin:
Can be closed.

PR #303, opened 2014-10-26
Remove init callback

xVin:
targets #298


matthiasbeyer:
Looks good to me. I don't see anything which is missing here.

neithernut:
If anything were missing the build would fail.
ACK.

Issue #267, opened 2014-10-22
Implement arithmetical commands


neithernut:
The stubs are in `src/commands/arithmetical.c`. The first argument should be set to the result.

xVin:
Please assign me if that's ok. I'm finally done with my presentation & ready to get back to work.

matthiasbeyer:
@xVin The string is more critical atm.

xVin:
String testing? Ya I thought I'd start doing this tomorrow & just wanted something small for today because I'm so tired.
If you refer to the string API commands, just assign me to that instead.

xVin:
Ok, just read the other thread. I'll try to fix the string stuff first then.

xVin:
As I can't work on the ws_string tests atm because of a problem with my build env (and the "unlock mystery"), I will do this first.

PR #304, opened 2014-10-26
Add arithmetical commands

matthiasbeyer:
Targets #267

---

I know that this PR will start a shitstorm, even if I tell you:
Calm down before reviewing this! I didn't do that to piss you
off or something!

---

So, here is why I wrote these implementations the way I did:
1. The code in `do_integral_arithmetic()` would be duplicated
   four times when writing it the other way.
2. I didn't use macros. Macros would be the alternative to this,
   but I know that some of you guys hate large macro foo, and this
   would clearly end up beeing a big shitload of macro foo!
3. I wonder how this will get optimized and whether the compiler
   is intelligent enough to optimize this into single functions.
   Probably it is not, but maybe it does some crazy optimization
   here we don't know yet.

---

I'm not pissed off or something if this gets rejected, but I
would like to have a calm and premeditated discussion about it,
without bad words and the like!

### DIFF
--- a/src/command/arithmetical.c
+++ b/src/command/arithmetical.c
@@ -91,7 +91,22 @@ int
 ws_builtin_cmd_mul(
     union ws_value_union* args
 ) {
-    //!< @todo: implement
+    intmax_t prod = 1;
+    intmax_t val;
+    union ws_value_union* it;
+
+    // iterate over all the arguments, checking whether they are ints
+    ITERATE_ARGS_TYPE(it, args, val, int) {
+        prod *= val;
+    }
+
+    if (!AT_END(it)) {
+        // Arrr! We must've hit something, captn!
+        return -EINVAL;
+    }
+
+    ws_value_union_reinit(args, WS_VALUE_TYPE_INT);
+    ws_value_int_set(&args->int_, prod);
     return 0;
 }
 
--- a/src/command/arithmetical.c
+++ b/src/command/arithmetical.c
@@ -114,6 +114,23 @@ int
 ws_builtin_cmd_div(
     union ws_value_union* args
 ) {
-    //!< @todo: implement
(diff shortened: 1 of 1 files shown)


src/command/arithmetical.c
```diff
@@ -25,37 +25,156 @@
...
+    while (ws_value_get_type(&args->value) != WS_VALUE_TYPE_NONE) {
+        if (ws_value_get_type(&args->value) != WS_VALUE_TYPE_INT) {
+            return -EINVAL;
+        }
+
+        intmax_t a = ws_value_int_get(&base->int_);
+        intmax_t b = ws_value_int_get(&args->int_);
+
+        ws_value_int_set(&base->int_, operation(a, b));
+    }
```
neithernut, line 124: I think you meant to iterate here...

neithernut:
This setup doesn't make sense for divisions. Divisions only make sense for _two_ arguments. For subtractions, it's a different matter, because they are often presented as sums of negative numbers.

---

**If** you feel that it makes sense to move iteration logic to another function (in a _fixed_ set of functions) , then _why_ didn't you just use an enumeration?
Franky, I think it makes sense to have a macro for iterations. If would have the obvious benefit that it may be used in other places too. In this case, a macro would be the more useful and elegant.
I'll write such a macro and push a PR.

src/command/arithmetical.c
matthiasbeyer, line 124: Yep.

matthiasbeyer:
I like your suggestion!

matthiasbeyer:
Fixed. I implemented the div directly.

neithernut:
What do you intend to do wthi this branch now? You can either completely rewrite it or dump it.

matthiasbeyer:
I fixed the division, the rest should work. I will of coures rewrite the helper function to use the new macros.

Issue #182, opened 2014-10-12
Add testing: ws_string type


matthiasbeyer:
The string type should be tested.

There should be unicode-tests included in this one.

xVin:
Another string issue :o
Can I do this? :-)

neithernut:
Sure.

matthiasbeyer:
Prio changed to critical, as I need functionality from this in my PR #91 which seems to be broken in the string implementation.

xVin:
@matthiasbeyer I won't be able to work on this today, but I'll have a look at it tomorrow morning. Feel free to stop by at the foo pool so that we can discuss.

matthiasbeyer:
I implemented the string type tests. They are failing.

@xVin please fetch my branch "add-string_type_tests" from my repository and use it to implement the fixes. Send all patches together as PR to this repository.

xVin:
Ok, I'll do this starting tomorrow - after the dreaded presentation is over.

matthiasbeyer:
I'm currently doing some further investigation on this and well... the tests are awefully slow (I don't know why) and functions which should only return do not (I don't know why) and the tests fail at points where they shouldn't (and I don't know why).

So, this seems to be a bit more advanced topic here.

matthiasbeyer:
This stuff gets more confusing the more I dig into it.It seems to me that the `ws_string_init()` function segfaults _at return_ (not on the `return` statement, ~~but on the `ret` opcode~~ edit: actually the `leave` opcode causes the segfault). This is _really_ confusing right now.

(I commented out everything which had to do with threading (locks/locking) for the current debugging).

TheNeikos:
Sounds like a buffer overflow to me? Could that be the problem? A stack allocated string overflowing is deadly for the opcodes that follow.

matthiasbeyer:
Nope, I don't find any the like. I'm far beyond this level right now, on the registers level, but the `rbp` and `rsp` registers have the same values right after function enter and before `leave`.

neithernut:
I think I made _some_ progress, I pushed the stuff I did to `add-string_type_tests` in my repo, so you can review the changes I did and use it as a starting point for further investigation.

---

Maybe we should rip out the `is_utf8`-stuff completely. Conversion to raw format may as well be _assumed_ to return an utf8-string.

neithernut:
I got a fairly big part to work.
In the end it was just fixing subtile mistakes and carefully reading the documentation.

[...]

PR #305, opened 2014-10-26
Add string type tests

matthiasbeyer:
Targets #182

### DIFF
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -273,21 +273,27 @@ ws_string_raw(
 ){
     ws_object_lock_read(&self->obj);
 
-    if (!self->is_utf8) {
-        ws_object_unlock(&self->obj);
+    int32_t dest_len;
+    UErrorCode err = U_ZERO_ERROR;
+
+    (void) u_strToUTF8(NULL, 0, &dest_len, self->str, self->charcount, &err);
+    if ((err != U_BUFFER_OVERFLOW_ERROR) && U_FAILURE(err)) {
         return NULL;
     }
 
-    char* output = calloc(self->charcount + 1, sizeof(*output));
-    int32_t dest_len;
-    UErrorCode err;
+    char* output;
+    output = calloc(dest_len + 1, sizeof(*output)); // +1 => Nullbyte
+    if (!output) {
+        return NULL;
+    }
 
+    err = U_ZERO_ERROR;
     output = u_strToUTF8(output, self->charcount, &dest_len, self->str,
                          self->charcount, &err);
 
     ws_object_unlock(&self->obj);
 
-    if (U_FAILURE(err)) {
+    if (U_FAILURE(err) || (dest_len <= 0)) {
         free(output);
         return NULL;
     }
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -312,17 +312,22 @@ ws_string_set_from_raw(
 
     UErrorCode err = U_ZERO_ERROR;
 
+    // get the length of the buffer to allocate
     int32_t len;
     (void) u_strFromUTF8(NULL, 0, &len, raw, -1, &err);
-    if (U_FAILURE(err)) {
+    if ((err != U_BUFFER_OVERFLOW_ERROR) && U_FAILURE(err)) {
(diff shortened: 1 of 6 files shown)


matthiasbeyer:
Cleanup still missing.

src/objects/string.c
```diff
@@ -274,20 +274,21 @@ ws_string_raw(
...
     ws_object_lock_read(&self->obj);
 
-    if (!self->is_utf8) {
+/*    if (!self->is_utf8) {
+        ws_object_unlock(&self->obj);
         return NULL;
-    }
+    }*/
 
     char* output = calloc(self->charcount + 1, sizeof(*output));
```
neithernut: The length should be calles as in `ws_string_set_from_raw()`.

tests/check/objects/ws_string/test.c
```diff
@@ -0,0 +1,269 @@
...
+    //tcase_add_test(test_string_multicat_empty);
+    tcase_add_test(tc, test_string_empty_dupl);
+    tcase_add_test(tc, test_string_empty_cmp);
+    tcase_add_test(tc, test_string_empty_raw);
+    tcase_add_test(tc, test_string_from_raw_literal);
+    tcase_add_test(tc, test_string_from_raw_heapbuf);
+
+//    tcase_add_test(tc_m, test_string_cmp);
+//    tcase_add_test(tc_m, test_string_ncmp);
+//    tcase_add_test(tc_m, test_string_substr);
```
neithernut: You realized that I commented these out just to suppress a bit of noise?

matthiasbeyer: Just noticed you commented out a bunch of stuff... I guess we can investigate together this evening?

matthiasbeyer:
Ready for review here. @neithernut , @TheNeikos

neithernut:
We have to decide what to do about the `utf8` flag. I'd throw it out completely.

matthiasbeyer:
I'm not the one to ask here as I do not really have a clue how this fits all together. We should support UTF-8 in the string type, I really don't care how it is implemented as long as it _is_ implemented.

neithernut:
The flag does nothing but track _whether_ the string is utf8. We can omit it completely and loose nothing. So just rip it out.

matthiasbeyer:
Can we do this in another PR? Would be a bit cleaner!

neithernut:
So you would rather keep commits which jus comment stuff out?

matthiasbeyer:
Nah, I'm a bit fucked up here... I will do it.

neithernut:
ACK for squashing.

matthiasbeyer:
Done. Feel free to suggest more.

src/objects/string.c
```diff
@@ -273,20 +270,16 @@ ws_string_raw(
 ){
     ws_object_lock_read(&self->obj);
 
-    if (!self->is_utf8) {
-        return NULL;
-    }
-
     char* output = calloc(self->charcount + 1, sizeof(*output));
```
neithernut: This `calloc()` call has yet to be fixed (see `ws_string_set_from_raw()`).

matthiasbeyer: What do you mean exactly?

neithernut: The number of units may differ from utf8 to the internal representation.

matthiasbeyer: So it should be a `UChar` instead of a `char`? (And a check is missing, wether the `calloc()` worked.=

neithernut: No, a pre-flight should be performed, like in `ws_string_set_from_raw()` **just as I pointed out already**.

neithernut:
https://github.com/matthiasbeyer/waysome/commit/0e3f5eb1b43d2d86e3197f0b72cf2e0a13862378 and https://github.com/matthiasbeyer/waysome/commit/53df29d502f7c131287a7678e35c86c4bff17abe should be squased into https://github.com/matthiasbeyer/waysome/commit/136e666c4b51a299d4ce23567215683037b1d8bc.

neithernut:
ACK.

PR #306, opened 2014-10-26
Add util header for subsystem specific macros

neithernut:
This PR adds utility macros for the command subsystem.

### DIFF
--- a/src/command/arithmetical.c
+++ b/src/command/arithmetical.c
@@ -25,13 +25,34 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
+#include <stdint.h>
+
 #include "command/arithmetical.h"
+#include "command/util.h"
+#include "values/union.h"
 
 int
 ws_builtin_cmd_add(
     union ws_value_union* args
 ) {
-    //!< @todo: implement
+
+    intmax_t sum = 0;
+    intmax_t val;
+    union ws_value_union* it;
+
+    // iterate over all the arguments, checking whether they are ints
+    ITERATE_ARGS_TYPE(it, args, val, int) {
+        sum += val;
+    }
+
+    if (!AT_END(it)) {
+        // Arrr! We must've hit something, captn!
+        return -EINVAL;
+    }
+
+    ws_value_union_reinit(args, WS_VALUE_TYPE_INT);
+    ws_value_int_set(&args->int_, sum);
     return 0;
 }
 
--- a/src/command/util.h
+++ b/src/command/util.h
@@ -48,5 +48,39 @@
 #define ITERATE_ARGS(it_, args_) \
     for (it_ = (args_); AT_END(it_); ++it_)
 
+
+/*
+ * Ugly block of macros for pure conversion, because _someone_ insisted on using
+ * uppercase names for his enum values...
+ */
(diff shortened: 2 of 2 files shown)


neithernut:
Note that the macros introduces by this PR are completely untested. I may (re)implement some command(s) to test them.

src/command/util.h
```diff
@@ -0,0 +1,85 @@
...
+ *
+ *      int foo;
+ *      ITERATE_ARGS(it, args, foo, int) {
+ *          // do stuff
+ *      }
+ */
+#define ITERATE_ARGS_TYPE(it_, args_, varname_, argtype_) \
+    for (union ws_value_union* it_ = (args); \
+         (ws_value_get_type(&it_->value) == WS_VALUE_TYPE_##argtype_ ) && \
+         (((varname_) = ws_value_##argtype_##_plain(it)), 1); \
```
matthiasbeyer: This does sadly only work for the boolean type. I guess you have to change the boolean type function to be named `get` here.

neithernut:
Waiting fpr #307.

src/command/util.h
```diff
@@ -0,0 +1,86 @@
...
+ *
+ *      int foo;
+ *      ITERATE_ARGS(it, args, foo, int) {
+ *          // do stuff
+ *      }
+ */
+#define ITERATE_ARGS_TYPE(it_, args_, varname_, type_) \
+    for (union ws_value_union* it_ = (args_); \
+         (ws_value_get_type(&it_->value) == WS_VALUE_TYPE_##type_ ) && \
+         (((varname_) = ws_value_##type_##_get((struct ws_value_##type_ *) \
```
TheNeikos, line 81: Jesus

matthiasbeyer, line 81: Of course. I would also stick to the uppercase names and call the macro with

```
ITERATE_ARGS(it, args, foo, INT)
```

Or something. Because when you have `object` instead of `int`, it looks a bit unfamiliar!

matthiasbeyer, line 81: As @neithernut pointed out, the uppercase stuff won't work. Hm. Well. Then, I guess, this is the best solution we can get without changing the styleguide.

matthiasbeyer:
Why is this on hold, btw?

neithernut:
Because I waited for #308.

matthiasbeyer:
Can you please test this macro? (Implement a function with it, compile it etc?) Just to be sure you don't have a minor bug in it? This hasn't to be included in this PR, of course.

matthiasbeyer:
Blocks #304

src/command/arithmetical.c
```diff
@@ -25,13 +25,34 @@
...
+    union ws_value_union* it;
+
+    // iterate over all the arguments, checking whether they are ints
+    ITERATE_ARGS_TYPE(it, args, val, int) {
+        sum += val;
+    }
+
+    if (!AT_END(it)) {
+        // Arrr! We must've hit something, captn!
+        return 1;
```
matthiasbeyer: Don't we return errno.h constants?

neithernut: Err... yeah. Was a demo anyways. I'll fix this.

neithernut:
The thing works.

matthiasbeyer:
ACK.

PR #311, opened 2014-10-27
Testing stuff

TheNeikos:


TheNeikos:
This was a test to see what kind of output drone gives in env variables

PR #315, opened 2014-10-27
Add input cleanup

TheNeikos:


src/input/module.c
```diff
@@ -120,17 +121,45 @@ find_initial_devices(void) {
 
         struct ws_input_device* new_dev = ws_input_device_new(fd);
 
-        ws_set_insert(&ws_input_ctx.devices, (struct ws_object*) new_dev);
+        ws_set_insert(&ws_input_ctx.devices,
+                        (struct ws_object*) getref(new_dev));
```
matthiasbeyer: The `getref()` macro casts to `struct ws_object*`, doesn't it? Or does it cast to input type?

src/input/module.c
```diff
@@ -120,17 +121,45 @@ find_initial_devices(void) {
...
     }
     closedir(d);
     return 0;
 }
 
+static void
+cleanup_input(
+    void* dummy
+) {
+
```
matthiasbeyer: You can remove this blank line.

src/input/module.c
```diff
@@ -120,17 +121,45 @@ find_initial_devices(void) {
...
+        ws_set_remove(&ws_input_ctx.devices, (struct ws_object*) dev);
+
+        libevdev_free(dev->dev);
+        ev_io_stop(ev_default_loop(EVFLAG_AUTO), &dev->watcher);
+        close(dev->fd);
+        ws_object_unref((struct ws_object*) dev);
+    }
+
+    ws_object_unref((struct ws_object*) &ws_input_ctx.devices);
```
matthiasbeyer: What type is the `ws_input_ctx.devices` variable? A `ws_object_unref()` causes a `free()` behind the scenes when the refcount gets zero, you know that?

TheNeikos: Shouldn't it only do that when the heapalloced setting is used?

matthiasbeyer: It should, but I'm not sure it behaves like we assume. :smile:

src/input/module.c
TheNeikos: [It casts to input type](https://github.com/waysome/waysome/blob/develop/src/objects/object.h#L56)

src/input/module.c
neithernut: `ws_object_unref()` _only_ calls `free()` if the heap-alloced-flag is set.
If if does, it's a bug.

matthiasbeyer: That's exactly what I said.

TheNeikos: You made it sound like if you were worried about freeing a stack reference.

The devices member is a ws_set.

neithernut:
ACK

matthiasbeyer:
ACK.

Btw, we have to stop this image posting foo. It was funny the first time, but now it's just traffic!

Issue #316, opened 2014-10-27
Make ws_object_unref() call the deinit callback even if it is allocated on the stack


neithernut:
Easy as cake.

PR #317, opened 2014-10-27
Moved check for heap allocation to end of unref

TheNeikos:
This is for #316


matthiasbeyer:
ACK.

TheNeikos:
Upon further reflection on how we designed the system, we don't think it is an appropriate change. Stack variables shouldn't be counted upon and only used in such a way that verifies their validity.

Thus you don't need refcounting and can deinit them right away!

Issue #313, opened 2014-10-27
Pass click-events to the currently active application.


neithernut:
This bit is missing. Once it's dome we've reached another milesone!

PR #319, opened 2014-10-27
Add mouse clicks

TheNeikos:


neithernut:
Targets #313.

src/input/input_device.c
```diff
@@ -136,6 +149,44 @@ cmp_callback(
...
 
+
+static void
+handle_relative_event(
+    struct input_event* ev
+) {
+    //ws_log(&log_ctx, LOG_DEBUG, "It's a mouse event!");
+
+    switch (ev->code) {
+        case REL_X:
```
matthiasbeyer: The case indentation level should be the same level as the switch.

matthiasbeyer:
ACK.

Issue #250, opened 2014-10-20
Add latex setup


neithernut:
We will need a setup to build the user documentation.

neithernut:
We'll use texinfo.

PR #320, opened 2014-10-27
Add build env for user manual

neithernut:
This PR introduces the initial setup for the user manual.
I did not yet decide in what section the manual should be put.
Also note that the documentation is _not_ build on the CI, since the setup omits building the documentation if either `latex2man` or `makeinfo` are missing.
Targets #250.


matthiasbeyer:
Looks good to me.

PR #333, opened 2014-10-28
Fix argument preparation in action processor

neithernut:
The action processor did use the new top for indirect arguments, when it should have used the old one.

### DIFF
--- a/src/action/processor_stack.c
+++ b/src/action/processor_stack.c
@@ -180,12 +180,20 @@ ws_processor_stack_bottom(
 struct ws_value*
 ws_processor_stack_value_at(
     struct ws_processor_stack* self,
-    ssize_t pos
+    ssize_t pos,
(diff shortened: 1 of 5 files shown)


src/action/processor_stack.c
```diff
@@ -180,12 +180,20 @@ ws_processor_stack_bottom(
 struct ws_value*
 ws_processor_stack_value_at(
     struct ws_processor_stack* self,
-    ssize_t pos
+    ssize_t pos,
+    struct ws_value* opttop
 ) {
+    size_t top = self->top;
+
+    // if `top` is `NULL`, we want to use the value we have
```
matthiasbeyer: `top` cannot be `NULL`, it is a `size_t`. You mean zero here?

matthiasbeyer:
ACK.

PR #336, opened 2014-10-28
Fix action processor

neithernut:
This PR brings various fixes for the command processor and ation processor.


neithernut:
One bit is still missing: The transaction is unref'd twice for some reason (both backtraces are identical).

---

I'll investigate this and notify you as soon as I've got a solution.

---

It was acutally a malformed loop. It's fixed now.

matthiasbeyer:
Looks good to me, but will test if you want me to.

PR #337, opened 2014-10-29
Add an empty config file to be found by waysome on startup

neithernut:
The configuration file may be expanded later.
Note that a few directories, including your home directory, will get searched first. This way you may still have waysome load your testing/personal configuration without "poluting" the `waysome.json` in the repo.


TheNeikos:
I agree with that.

ACK

matthiasbeyer:
Can we make it a hidden file?

neithernut:
Yes, `.waysome.json` would also be loaded. But why would you want to make it a hidden file?
For clarification: the config would be installed to `/etc/something`, not to `$HOME`.

matthiasbeyer:
Because it is added to the repo. 

What's the purpose of the config file in the repo? What's the purpose of it when it is empty? If the purpose is that you can test, I would like to suggest: Put an entry in the gitignore, so one can actually put stuff into it without checking it in.

matthiasbeyer:
Missed the description of the PR. ACK.

PR #341, opened 2014-10-29
Fix the stack top fetching in ws_processor_prepare_args()

matthiasbeyer:
This patch fixes the stack top fetching in the
ws_processor_prepare_args() function. The problem was that the pushing
to the stack changes the stack and _realloc()_s it, if required. That
causes the stack object to be located in another position, which caused
everything to fail.

Suggested-by: Julian Ganz julian.ganz@hs-furtwangen.de


neithernut:
ACK.

PR #343, opened 2014-10-29
Fix type of buffer var

neithernut:
If we internally use intmax_t, but the buffer used for transmission is
an uin32_t, we end up with completely different values for negative
and/or big numbers.


matthiasbeyer:
ACK.

PR #344, opened 2014-10-29
Fix: ws_value_int_get()

matthiasbeyer:
... to return the proper type and don't accidentially
cast away information.


neithernut:
ACK.

PR #345, opened 2014-10-29
Fix JSON deserializer for restarting parsing

matthiasbeyer:
This PR fixes the JSON deserializer so it is able to continue parsing.

### DIFF
--- a/src/serialize/json/deserializer.c
+++ b/src/serialize/json/deserializer.c
@@ -35,6 +35,7 @@
  * yajl library, see: https://lloyd.github.io/yajl/
  */
 
+#include <ctype.h>
 #include <string.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


src/serialize/json/deserializer.c
```diff
@@ -132,8 +133,18 @@ deserialize(
...
     struct deserializer_state* d = self->state;
+    size_t consumed = 0;
+
+    while (isspace(*buffer)) {
+        buffer++;
+        consumed++;
+        if (consumed == nbuf) {
+            self->is_ready = true;
+            return consumed;
+        }
```
neithernut, line 145: You could also decrement `nbuf` and check whether it's `0`. This way you even wouldn't have to subtract `consumed` from `nbuf` later.

neithernut:
ACK.

PR #347, opened 2014-10-29
Fixed erroneous return with a continue

TheNeikos:
This bug stopped the flushing of the event queue by returning too early


neithernut:
ACK.

Issue #350, opened 2014-10-29
ws_string must be initializeable from ws_string itself


matthiasbeyer:
The interface should look like this:

``` C
int
ws_string_init_from_str(
    struct ws_string* self,
    struct ws_string const* other
);
```

matthiasbeyer:
Blocks #323

PR #352, opened 2014-10-30
Add ws string init from str

xVin:
targets #350

### DIFF
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -90,6 +90,37 @@ ws_string_init(
     return false;
 }
 
+bool
+ws_string_set_from_str(
+    struct ws_string* self,
+    struct ws_string* other
+) {
+    if (!self || !other) {
+        return false;
+    }
+
+    ws_object_lock_write(&self->obj);
+    ws_object_lock_read(&other->obj);
+
+    UChar* temp;
+    temp = realloc(self->str, (other->charcount + 1) * sizeof(*self->str));
+    if (!temp) {
+        ws_object_unlock(&other->obj);
+        ws_object_unlock(&self->obj);
+        return false;
+    }
+    
+    self->str = temp;
+    self->charcount = other->charcount;
+
+    self->str = u_strcpy(self->str, other->str);
+
+    ws_object_unlock(&other->obj);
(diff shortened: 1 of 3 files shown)


src/objects/string.c
```diff
@@ -90,6 +90,34 @@ ws_string_init(
...
+
+    ws_object_unlock(&other->obj);
+
+    if (self->str) {
+        ws_object_unlock(&self->obj);
+        return true;
+    }
+
+    ws_object_unlock(&self->obj);
+    return false;
```
matthiasbeyer: You can put the lines 112 to 118 into two lines:

``` C
ws_object_unlick(&self->obj);
return !!self->str;
```

src/objects/string.h
```diff
@@ -62,7 +62,7 @@ extern ws_object_type_id WS_OBJECT_TYPE_ID_STRING;
  * @memberof ws_string
  *
  * @return true if ws_string object was successfully initialized,
- * NULL on failire
+ * NULL on failure
```
matthiasbeyer: This is false anyway, a boolean cannot be `NULL`.

xVin:
I added another commit to avoid code duplication.

src/objects/string.c
```diff
@@ -90,6 +90,29 @@ ws_string_init(
...
+    if (!self || !other) {
+        return false;
+    }
+
+    ws_object_lock_write(&self->obj);
+    ws_object_lock_read(&other->obj);
+
+    self->charcount = other->charcount;
+
+    self->str = realloc(self->str, (self->charcount + 1) * sizeof(*self->str));
```
matthiasbeyer: You lose `self-str` here when `realloc()` returns NULL.

src/objects/string.c
```diff
@@ -90,6 +90,29 @@ ws_string_init(
...
+    self->str = u_strcpy(self->str, other->str);
```
matthiasbeyer, line 116: Same problem here.

matthiasbeyer:
ACK.

neithernut:
ACK.

matthiasbeyer:
Jup.

PR #353, opened 2014-10-30
Fix ws string reallocs

xVin:
Includes guards in case realloc fails.

### DIFF
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -202,7 +202,14 @@ ws_string_multicat(
         }
     }
 
-    self->str = realloc(self->str, (len + 1) * sizeof(*self->str));
+    UChar* temp;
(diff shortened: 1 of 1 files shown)


src/objects/string.c
```diff
@@ -157,8 +157,16 @@ ws_string_cat(
     ws_object_lock_write(&self->obj);
     ws_object_lock_read(&other->obj); //!< @todo Thread-safeness!
 
-    self->str = realloc(self->str, (self->charcount + other->charcount + 1)
+    UChar* temp;
+    temp = realloc(self->str, (self->charcount + other->charcount + 1)
                         * sizeof(*self->str));
```
neithernut: I know _someone_ who will say that the indendation has to be fixed here.

matthiasbeyer:
I think this looks good so far.

neithernut:
ACK.

Issue #322, opened 2014-10-27
Add set holding transactions to be triggered by an event


neithernut:
The transactions also have to be triggerable by events. Note that a transaction also has an id and a name. We may adjust the hash and the cmp callback to enable searching for a transaction.
Additionally, we also need some mapping from events to transactions.

Part of a PR targeting this would also be adding transactions if the appropriate flag is set and reacting to events in `ws_action_manager_process()`.

neithernut:
I have to do this myself.

PR #354, opened 2014-10-30
Add event-transaction mapping in action manager

matthiasbeyer:
This is WIP, but I like to read suggestions. It adds support for
mapping transactions on events (by name) inside the action manager.

### DIFF
--- a/src/action/manager.c
+++ b/src/action/manager.c
@@ -36,6 +36,7 @@
 #include "objects/message/value_reply.h"
 #include "objects/set.h"
 #include "objects/string.h"
+#include "util/condition.h"
 
 /**
  * Event-transaction mapping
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Targets #322 btw.

neithernut:
I'll have a look at it later.

neithernut:
This will not work, since `ws_set` will call `ws_object` methods on the struct, which, in turn, used the type information within the struct. This will cause the entire thing to happily segfault or worse (e.g. illegal instruction).
Wouldn't it make more sense to have the `ws_event` type, derive some type which has a list of transactions and use a comparision which is based on the event's name?

---

This may only work if `ws_object_cmp` allowes comparing two objects of which one's type is derived from the other's, so maybe we want to extend _that_ too.

matthiasbeyer:
You did read my comment for the first member? `ws_set` will call `ws_object` methods, but it will automatically call these on the `ws_string` member, which is a `ws_object` itself.

Maybe I'm getting something wrong here, but the way it is it should work, as the `ws_string` member is embedded into the struct directly.

neithernut:
Ah, now I get it. I didn't read carefully enough.

matthiasbeyer:
:smiley:

neithernut:
**Please** make the context a static anonymous struct like in **all** the other modules. Manually allocating this is completely bogous.

neithernut:
ACK for squashing.

matthiasbeyer:
No need. 6 Commits is few enough, I guess. Feel free to suggest squashes, but I don't see need of any squashing here.

neithernut:
I think the last two commits can be squashed.

neithernut:
ACK.

neithernut:
This was clearly merged too early. I'll have to rewrite parts of it.

Issue #358, opened 2014-10-31
Add ws_event_new


neithernut:
Should do nothing but allocate memory, call `ws_event_init()` and set the `HEAPALLOCED` flag.
Will need this for my keyboard stuff, but it belong's in another PR.

PR #359, opened 2014-10-31
[WIP] Add ws event new

xVin:
targets #358

### DIFF
--- a/src/objects/message/event.c
+++ b/src/objects/message/event.c
@@ -104,6 +104,28 @@ out:
     return ret;
 }
 
+struct ws_event*
+ws_event_new(
+    struct ws_string* name,
+    struct ws_value* ctx
+) {
+    struct ws_event* ev = calloc(1, sizeof(*ev));
+
+    if (!ev) {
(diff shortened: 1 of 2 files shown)


xVin:
Is this what you had in mind?

neithernut:
Yep. Pretty much.
ACK.

src/objects/message/event.c
```diff
@@ -104,6 +104,28 @@ ws_event_init(
...
+    struct ws_event* ev = calloc(1, sizeof(*ev));
+
+    if (!ev) {
+        return NULL;
+    }
+
+    int res = ws_event_init(ev, name, ctx);
+
+    if (res != 0) {
+        return NULL;
```
matthiasbeyer, line 121: Does not `free()` the allocated memory here.

Maybe you should provide a follow-up patch to fix this.

xVin, line 121: True & done.

neithernut:
Argh, I should stop merging PRs at 3am.

Issue #270, opened 2014-10-22
Add string commands


neithernut:
We need some commands for strings in the API.

PR #362, opened 2014-11-01
[WIP] Add string commands

xVin:
targets #270

### DIFF
--- a/src/command/string.c
+++ b/src/command/string.c
@@ -1,0 +1,85 @@
+/* waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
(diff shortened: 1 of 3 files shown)


src/command/string.c
```diff
@@ -0,0 +1,137 @@
...
+    struct ws_string* str2;
+
+    int res;
+    res = get_strings_from_union(args, &str1, &str2);
+
+    if (res != 0) {
+        return res;
+    }
+
+    bool res2 = ws_string_substr(str1, str2);
```
matthiasbeyer: I would call this variable `issubstr`.

src/command/string.c
```diff
@@ -0,0 +1,137 @@
...
+    struct ws_string** str1,
+    struct ws_string** str2
+) {
+    if (ws_value_get_type(&args->value) != WS_VALUE_TYPE_STRING ||
+            ws_value_get_type(&args[1].value) != WS_VALUE_TYPE_STRING) {
+        return -EINVAL;
+    }
+
+    if (ws_value_get_type(&args[2].value) != WS_VALUE_TYPE_NONE) {
+    return -E2BIG;
```
matthiasbeyer: Indentation!

xVin:
Please review & tell me if we need more string commands.
Should we split strcat into strcat and strncat?

neithernut:
Nah. It's more intuitive the way it is.

src/command/string.c
```diff
@@ -0,0 +1,154 @@
...
+    union ws_value_union* args
+) {
+    union ws_value_union* it;
+    struct ws_string* val;
+    struct ws_string* res = args->string.str;
+
+    //iterate over all arguments, checking whether they are ws_value_strings
+    ITERATE_ARGS_TYPE(it, args, val, string) {
+        res = ws_string_cat(res, val);
+    }
```
neithernut, line 73: Argh, I removed @matthiasbeyer's comment because I got confused about where this scope ends.
He was right when he said that `res` might be NULL. You should return/break in this case.

xVin, line 73: As you can see a few lines below, I already fixed this.

neithernut, line 73: Yes, that's what I thought, but you have to resolve this within the loop.
If `ws_string_cat()` returns `NULL`, you end up passing it to `ws_string_cat()` in the next iteration.

xVin, line 73: Ah, yes, of course <__<

matthiasbeyer:
Looks good for me, @neithernut ?

neithernut:
The one thing I annotated was fixed.
ACK.

PR #370, opened 2014-11-04
Add missing member NULLing for (hopefully) all types

matthiasbeyer:
This PR adds member NULLing for (hopefully) all types where the
type id table was not yet updated.

Please revewi and suggest more position I may have missed.


neithernut:
ACK.

PR #373, opened 2014-11-04
Add action manager transaction storage

neithernut:
This PR adds a transaction storage, a transaction registration. It also adds code which let's you trigger transactions through events.
Targets #322.

### DIFF
--- a/src/action/manager.c
+++ b/src/action/manager.c
@@ -55,7 +55,8 @@
  */
 static struct ws_reply*
 run_transaction(
-    struct ws_transaction* transaction // transaction to run
+    struct ws_transaction* transaction, // transaction to run
+    struct ws_value* context //!< context to push on the stack
 );
 
 
@@ -81,7 +82,7 @@ ws_action_manager_process(
 
         if (flags & WS_TRANSACTION_FLAGS_EXEC) {
             // execute the transaction
-            return run_transaction(transaction);
+            return run_transaction(transaction, NULL);
         }
 
         return NULL;
@@ -101,7 +102,8 @@ ws_action_manager_process(
 
 static struct ws_reply*
 run_transaction(
-    struct ws_transaction* transaction // transaction to run
+    struct ws_transaction* transaction,
+    struct ws_value* context
 ) {
     struct ws_reply* retval = NULL;
     int res;
@@ -116,7 +118,29 @@ run_transaction(
         goto cleanup_stack;
     }
 
-    //!< @todo push environment on the stack
+    // push environment on the stack
+    res = ws_processor_stack_push(&stack, 2);
+    if (res < 0) {
+        retval = (struct ws_reply*)
+                 ws_error_reply_new(transaction, -res, "Could not init stack",
+                                    NULL);
+        goto cleanup_stack;
+    }
+
+    {
+        union ws_value_union* bottom = ws_processor_stack_bottom(&stack);
+
+        // initialize the global context
+        ws_value_union_reinit(bottom, WS_VALUE_TYPE_NIL); //!< @todo real thing
+
+        // initialize the event context
+        ++bottom;
+        if (context) {
+            ws_value_union_init_from_val(bottom, context);
+        } else {
+            ws_value_union_reinit(bottom, WS_VALUE_TYPE_NIL);
+        }
+    }
 
     // we start a new frame, but we will never restore the default frame
     (void) ws_processor_stack_start_frame(&stack);
--- a/src/action/manager.c
+++ b/src/action/manager.c
@@ -194,6 +194,44 @@ ws_action_manager_process(
     return NULL;
 }
 
+int
+ws_action_manager_register(
+    struct ws_string* event_name,
+    struct ws_string* transaction_name
+) {
+    int res;
+
+    // get the transaction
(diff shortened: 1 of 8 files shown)


src/action/manager.c
```diff
@@ -65,6 +87,34 @@ run_transaction(
...
+    }
+
+    res = ws_set_init(&actman_ctx.registrations);
+    if (res < 0) {
+        goto cleanup_transactions;
+    }
+
+    ws_cleaner_add(action_manager_deinit, NULL);
+
+    is_init= true;
```
matthiasbeyer: Missing space around `=`

src/action/manager.c
```diff
@@ -77,21 +127,82 @@ ws_action_manager_process(
...
+            named = set_get(&actman_ctx.registrations, &comparable);
+
+            ws_object_deinit((struct ws_object*) &comparable);
+
+            // extract the transaction to run
+            if (!named) {
+                return NULL;
+            }
+
+            transaction = (struct ws_transaction*) ws_named_get_obj(named);
```
matthiasbeyer, line 177: Where is the object inside the named value set? I don't see where the transaction object is put into the named object, please point me to it!

src/action/manager.c
```diff
@@ -116,7 +228,29 @@ run_transaction(
...
+
+        // initialize the global context
+        ws_value_union_reinit(bottom, WS_VALUE_TYPE_NIL); //!< @todo real thing
+
+        // initialize the event context
+        ++bottom;
+        if (context) {
+            ws_value_union_init_from_val(bottom, context);
+        } else {
+            ws_value_union_reinit(bottom, WS_VALUE_TYPE_NIL);
```
matthiasbeyer, line 324: Wouldn't `--bottom` be enough here, as you already initialized one bottom stack position with NIL?

src/objects/named.h
```diff
@@ -0,0 +1,102 @@
...
+};
+
+/**
+* ws_named type information
+*/
+extern ws_object_type_id WS_OBJECT_TYPE_ID_NAMED;
+
+/**
+ * Initialize a new named object
+ */
```
matthiasbeyer, line 67: `@memberof` and `@return` missing.

src/objects/named.h
```diff
@@ -0,0 +1,102 @@
...
+    struct ws_named* self, //!< named value to initialize
+    struct ws_string* name, //!< name of the object
+    struct ws_object* val //!< valu eof the object
+)
+__ws_nonnull__(1)
+;
+
+/**
+ * Create a new named object
+ */
```
matthiasbeyer, line 83: `@memberof` and `@return` missing.

src/objects/named.h
```diff
@@ -0,0 +1,102 @@
...
+ws_named_new(
+    struct ws_string* name, //!< name of the object
+    struct ws_object* val //!< valu eof the object
+)
+__ws_nonnull__(1)
+;
+
+/**
+ * Get a reference of the object
+ */
```
matthiasbeyer, line 98: `@memberof` and `@return` missing.

src/action/manager.c
neithernut, line 177: In `ws_action_manager_register()`.

src/action/manager.c
neithernut, line 324: wtf are you talking about?

neithernut:
Note: with this implementation, it's an event can only trigger one transaction. This may, however, be changed in the future (after this PR). It's fairly easy to do it.

src/action/manager.c
matthiasbeyer, line 324: You are doing a initialization of the bottom value in the else case here, I _guess_ (I don't know) this is because the bottom of a stack must be NIL to be able to iterate over the stack values inside a function or whatever and stopping when hitting NIL.

But you already added a NIL value on the stack before the whole if-else construct, so you're basically have two NIL values on the stack after the else branch here.

My question is: It is necessary to have two NIL values on the stack, wouldn't be one enough? Or would this break other parts of the code if one NIL value would be on the stack when there is no context around here?

neithernut, line 324: What you're talking is, basically, bullshit.
You have _two_ contexts (a global one and an event context), which are addressable statically by the hardcoded stack positions `0` and `1` (as described by the comments _just_ above the code).
A nil value just indicates that the context is _not_ set.

To indicate the end of a list, you'd use an invalid value ("none") _anyways_.

matthiasbeyer, line 324: Can you _**PLEASE**_ stop saying "bullshit" all the time.

I'm asking normal questions here, to be able to understand _why_ you do things as you do them. So nothing of the said is bullshit, it's a question if it could be done this way too and if not, why. I'm asking to _understand_, not to annoy! You always using terms like "bullshit" and others make people (me) uncomfortable and annoyed by _you_. I suggest you take a break for one day and get some sleep. Because _you_ are annyoing _me_ here with your aggressive way of answering on questions.

matthiasbeyer:
ACK.

src/action/manager.c
neithernut, line 324: I had enough of a break today as I was unable to progress any further with my tasks.

Issue #180, opened 2014-10-12
Add testing: ws_array type


matthiasbeyer:
Testing for the array type should be added.

matthiasbeyer:
@neithernut , @TheNeikos Do we actually need the array type? I wouldn't put effort into this if the array types gets removed anyway!

neithernut:
Mh... I will check this when I'm back in FuWa.

neithernut:
The array type can be dumped.

matthiasbeyer:
I will throw this (ws_array) out now.

neithernut:
The type was removed. Testing is thus not an issue we have to resolve.

PR #379, opened 2014-11-06
Remove ws_array type

matthiasbeyer:
Targets (partially) #180

The ws_array type was not used except for a function of ws_string, which
was unused. I removed the `ws_string_multicat()` function, fixed the C/P
failures which made tools like `ack` say that there is usage of `ws_array`,
and finally removed the `ws_array` implementations itself.


neithernut:
ACK.

PR #383, opened 2014-11-07
Changed the deinit call to an unref to prevent a wrong free

TheNeikos:
This fixes a free of a stackalloced value


matthiasbeyer:
Tested. ACK.

Issue #385, opened 2014-11-07
`ws_string_len()` seems to return the length of the string + 1


matthiasbeyer:
If you look at [this](https://github.com/waysome/waysome/pull/357/files#diff-8a6325a071ab3a6eadc10be77d828dc1R332) line, which is run in [this test](https://github.com/waysome/waysome/pull/357/files#diff-cc90077ce3da298b7378c5d67e872012R116) with a string which is 9 characters (10 byte incl. the `NULL`-Byte), it returns `11` instead of `10`.

This should be fixed, as it blocks my testing in #357 (I need to test against `"teststring\\u000"` instead of `"teststring"`, which is ugly and shouldn't be required).

PR #386, opened 2014-11-07
Fix: string length issue

matthiasbeyer:
This addresses #385.

### DIFF
--- a/src/objects/string.c
+++ b/src/objects/string.c
@@ -143,7 +143,7 @@ ws_string_len(
     if (self) {
         size_t len;
         ws_object_lock_read(&self->obj);
-        len = self->charcount;
+        len = u_strlen(self->str);
         ws_object_unlock(&self->obj);
         return len;
     }
--- a/src/objects/string.h
+++ b/src/objects/string.h
@@ -93,7 +93,7 @@ struct ws_string*
 ws_string_new(void);
 
 /**
- * Get length of a ws_string
+ * Get length of a ws_string in number of characters
  *
  * @memberof ws_string
  *
--- a/tests/check/objects/ws_string/test.c
+++ b/tests/check/objects/ws_string/test.c
@@ -151,6 +151,9 @@ START_TEST (test_string_from_raw_literal) {
     ck_assert(buf != NULL);
     ck_assert(0 == strcmp(buf, literal));
 
+    size_t len = ws_string_len(s);
+    ck_assert(strlen(literal) == len);
+
     ws_object_deinit(&s->obj);
     free(s);
 }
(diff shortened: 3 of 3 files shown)


matthiasbeyer:
Ready for review and merge btw.

matthiasbeyer:
Well, it actually should return `9` for `9` characters in a string, so this is PR not complete by now. I will fix this tomorrow!

neithernut:
So the length set by `u_strFromUTF8` _does_ contain the termination character after all...
Would suggest fixing this by subtracting `1` from the length before returning it in `ws_string_len()`.

TheNeikos:
Please, please please please get up to date why we would use UTF-8 and what it means to support UTF-8!

First of all, UTF 8 is used to support Unicode, else we would just support ASCII (which would be easier, but we want dat ☺☻♥♦♣ etc...) Now, using `strlen` on an UTF-8 string is only calling for trouble! 

Sure, if you only use ASCII values it works _only_ because UTF-8 is meant to be backwards compatible to ASCII. Now, let's talk about a real use case, writing my name for example. 

`Marcel Müller` `strlen()` will report 14! If you don't want to count, there are actually only 13 characters. I am sure you can guess by now why, UTF-8 is meant to be multibyte! And ü is encoded in two bytes,  `0xC3` and `0xBC` to be exact, and they cannot be represented in ascii, as they are bigger than `0x7F`! However, as they are still valid bytes, strlen counts them as two.

I can explain this further if you want.

``` c
#include <stdio.h>
#include <string.h>

main()
{
   printf("%d =/= %d", strlen("Marcel Müller"), strlen("Marcel Muller"));
}
```

TheNeikos:
Also, I want to point to all the security holes that exist due to Unicode and UTF-8 and similar.

So many internet sites are vulnerable to supporting UTF-8 in their usernames for examples. BUT if your database for example doesn't support it you have interesting collisions. Same thing for buffer overflows, if you were to allocate a char array by using the char count utility given by multibyte string libraries you /will/ open all gates to hell as one could just insert arbitrary code into it.

[Just one of the things that can go wrong with Unicode](https://labs.spotify.com/2013/06/18/creative-usernames/) while it is _not_ as critical as for Spotify. Remember that waysome will run at a somewhat elevated position, the last thing you want is a security hole in the string component.

matthiasbeyer:
It seems that we do not need the `charcount` variable at all. Should we remove them?

neithernut:
@TheNeikos made some good point, which I totally forgot about at the moment I wrote my last comment.
Yes, we should definitely remove the `charcount` variable.

xVin:
I'm gonna do this later today.

matthiasbeyer:
@neithernut Please review.

neithernut:
ACK.

TheNeikos:
I still don't agree with the use of strlen in the tests. You should have also added a test which uses unicode characters as I had advised in my previous comment.

neithernut:
ACK, the tests still have to be fixed, but the underlying issue in the string class is.

Issue #351, opened 2014-10-30
Drop info pages?


neithernut:
The current setup is aimed at a info page and a man page, both generated from the same/similar sources. Now it turned out that the `latex2man` utility is not as usable as one may whish. And I, for my part, would not like to start writing `texinfo` documentation. (it's... ugly)

If we were to drop the info pages, there would be many options. E.g. we could generate the suer documentation in PDF and HTML format (using [TTH](http://hutchinson.belmont.ma.us/tth/), [tex2page](http://www.ccs.neu.edu/home/dorai/tex2page/index.html) or [another tool](http://ctan.org/topic/cvt-html)).
The man page could still be generated form a restricted subset of the sources for the manual (from a different main file) using `latex2man` or hand-crafted. Despite the problems we had with `latex2man`, I think the former is still a viable option (as `latex2man` is clearly meant to produce man pages).

matthiasbeyer:
To get this discussion going: ACK. I would rather provide an extensive manpage then a crappy infopage and a short manpage.

Others do it this way, too, just look at the manpages of git. So I don't see any problem with a large manpage.

PR #387, opened 2014-11-08
Produce a pdf rahter than an info page

neithernut:
Targets #351.

The initial setup may still lack a view bits (e.g. inclusion of LaTeX packages).


CMakeLists.txt
```diff
@@ -24,19 +24,20 @@ find_package(EGL REQUIRED)
 find_package(EV REQUIRED)
 find_package(IcuIO REQUIRED)
 find_package(IcuUC REQUIRED)
-find_package(libreset REQUIRED)
+find_package(LATEX REQUIRED)
```
matthiasbeyer: You would make the latex package required for building the software? If I understand that right: You actually need the latex package for building the software itself?

Can we define this to be required for the `doc` target only?

neithernut: What's really required here is the `FindLATEX` module, which is provided by the cmake installation. That module itself sets a few variables (e.g. `PDFLATEX_COMPILER`).
Requiring this package does _not_ mean we require `pdflatex` or anything latex related _at all_.

---

This would be more obvious if the PR was processed by the CI.

matthiasbeyer: Okay, got it.

Strange that the CI does not build it.

matthiasbeyer: It actually [compiled it](http://ci.neikos.me/github.com/waysome/waysome/develop/d6649c1699b220c0c895886a35eb97749a8adb3d) but it wasn't linked ... strange!

TheNeikos: That can happen sometimes when a packet gets lost I suppose.

neithernut: Anyways, I will remove the `REQUIRED` flag. It isn't exactely _neccessary_ and it _may_ be easier to understand that `latex` isn't a hard requirement for building `waysome` this way.

neithernut:
I will modify the build env to support out-of-source builds.

neithernut:
This _should_ now be ready for merge. Please review and notify me if anything (e.g. packages) are missing.

doc/user/CMakeLists.txt
```diff
@@ -3,27 +3,55 @@
...
 )
 
-add_custom_command(OUTPUT waysome.info
-        COMMAND ${MAKEINFO_EXECUTABLE} waysome.texi
-    MAIN_DEPENDENCY waysome.texi
+set(MANPAGE_TEX ${CMAKE_CURRENT_SOURCE_DIR}/manpage.tex)
+
+
+#
+# contrary to otehr compilers, pdflatex doesn't like the sources to be in
```
matthiasbeyer: Typo

doc/user/waysome.tex
```diff
@@ -0,0 +1,20 @@
+\documentclass{book}
+
+%
+% Packages
+%
+\usepackage[english]{babel}
+\usepackage[T1]{fontenc}
+\usepackage[utf8]{inputenc}
+\usepackage{listings}
+
```
matthiasbeyer, line 14: I guess we should also add these packages
-  `color`
- `url`
- `hyperref`

matthiasbeyer:
ACK.

PR #388, opened 2014-11-09
Remove charcount from string

xVin:


matthiasbeyer:
ACK from me.

neithernut:
I just read up on `u_strlen()`, just to be absolutely sure this works as intended. ACK.

PR #357, opened 2014-10-31
Add JSON serializer

matthiasbeyer:
This is WIP and targets #355.

PR #390, opened 2014-11-10
Add ws_value_set_get() function

matthiasbeyer:
Required for my work on #357.

The renaming of the old function was done, so the `ws_value_set_get()`
function name is consistent with the corresponding functions of the other
value types.

Blocks #357


neithernut:
ACK.

Issue #396, opened 2014-11-11
Add code for processing of (potential) key-combinations


neithernut:
I already have code for this one.

neithernut:
Blocks #400.

PR #397, opened 2014-11-11
Add hotkey evaluation

neithernut:
Targets #396.
I intend to rewrite the key caching and playback, so this is WIP.

### DIFF
--- a/src/input/hotkey_event.c
+++ b/src/input/hotkey_event.c
@@ -25,6 +25,7 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
 #include <malloc.h>
 
 #include "input/hotkey_event.h"
@@ -78,8 +79,9 @@ ws_object_type_id WS_OBJECT_TYPE_ID_HOTKEY_EVENT = {
  *
  */
 
-struct ws_hotkey_event*
-ws_hotkey_event_new(
+int
+ws_hotkey_event_init(
+    struct ws_hotkey_event* self,
     struct ws_string* name,
     uint16_t* codes,
     uint16_t code_num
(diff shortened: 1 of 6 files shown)


neithernut:
After I fixed the bug I just introduced, this is ready for review and, eventually, merge.

---

I found the source of the bug. It was an embarrasing one, and a nasty one to track down, too.

neithernut:
I'll bring in a minor optimization now.

---

I optimized the bit and also added some missing documentation.

src/input/hotkey_event.c
```diff
@@ -94,38 +96,56 @@ ws_hotkey_event_new(
...
             }
         }
     }
 
-    struct ws_hotkey_event* retval;
-    retval = (struct ws_hotkey_event*) ws_object_new(sizeof(*retval));
-    if (!retval) {
-        return NULL;
+    if (!ws_object_init(&self->obj)) {
+        return -1;
```
matthiasbeyer, line 106: Is there any reason why you do not return the return value of the init call?

src/input/hotkey_event.c
```diff
@@ -94,38 +96,56 @@ ws_hotkey_event_new(
...
 cleanup_string:
-    ws_object_deinit(&retval->name.obj);
-
+    ws_object_deinit(&self->name.obj);
 cleanup_object:
-    ws_object_deinit(&retval->obj);
-    free(retval);
-    return NULL;
+    ws_object_deinit(&self->obj);
+    return -1;
```
matthiasbeyer, line 127: Is there any reason why you do not return a errno.h number?

neithernut, line 127: Because I don't get any useful info _why_ the functions which failed failed.

src/input/hotkey_event.c
neithernut, line 106: Because it's a `bool`.

src/input/hotkeys.c
```diff
@@ -0,0 +1,256 @@
...
+        // jump to the next DAG node
+        ws_hotkeys_ctx.state = ws_hotkey_dag_next(ws_hotkeys_ctx.state, code);
+        if (!ws_hotkeys_ctx.state) {
+            // this key is not part of a keycombo
+            ws_log(&ws_hotkeys_ctx.log, LOG_DEBUG,
+                   "Keycode %d not part of key-combo", code);
+            return eventlist_reset();
+        }
+
+        ws_log(&ws_hotkeys_ctx.log, LOG_INFO,
```
matthiasbeyer, line 145: I guess we should add a macro for debugging calls like this, to exclude calls to the log function when building without debugging. I guess this would be beneficial in cases like this.

Shall I add this?

neithernut, line 145: Noin this branch. Open an issue.

matthiasbeyer:
ACK & Merge.

PR #398, opened 2014-11-11
Remove all the dependencies we don't need

neithernut:
The problem which is solved by this PR is a general one, and the build env will be optimized at some point to prevent this happening again.
However, **tests should not pull in all the modules as dependencies!**


matthiasbeyer:
ACK.

Issue #326, opened 2014-10-28
Structure user documentation


neithernut:
The user documentation in `doc/user` should be extended from, well... nothing, to a structured document. This PR is purely about structure. Content is not neccessary.

neithernut:
Blocks #327.

PR #403, opened 2014-11-14
Add structure for user documentation

matthiasbeyer:
Targets #326

### DIFF
--- a/doc/user/waysome.tex
+++ b/doc/user/waysome.tex
@@ -18,7 +18,25 @@
 %
 \begin{document}
 
-%TODO: structure
+% Import the title page
+%\section{title} -- No title section here
+\input{src/title}
+
+% Import the description of waysome
+\section{Description}
+\input{src/description}
+
+% Import the environment description of waysome
+\section{Environment}
+\input{src/environment}
+
+% Import the file descriptions of waysome
+\section{Files}
+\input{src/files}
+
+% Import the API description - what this all is about
+\section{API}
+\input{src/api}
 
 \end{document}
 
--- a/doc/user/manpage.tex
+++ b/doc/user/manpage.tex
@@ -1,0 +1,46 @@
+\documentclass{book}
+
+%
+% Packages
+%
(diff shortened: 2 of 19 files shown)


neithernut:
Sections missing:
- [X] Command invocation
- [X] Files (socket etc)
  - [X] Socket(s)
  - [X] Config file (search order)
- [X] Environment

matthiasbeyer:
I added the environment and the file files, the latter in one file by now.

I would include the command invocation in the protocol part / API description.

neithernut:
By "command invocation" I do **not** mean ivoking API commands, but invoking `waysome` itself.

matthiasbeyer:
Ah, okay. Is it necessary to have a dedicated section for this? Could it go into the "description" part of the documentation?

neithernut:
Nope, the conventional manpage format requires a section about the invocation.

matthiasbeyer:
Ok. I've added it.

doc/user/waysome.tex
```diff
@@ -18,7 +18,23 @@
...
+% Import the title page
+\input{src/invocation}
+
+% Import the description of waysome
+\input{src/description}
+
+% Import the description of waysome
+\input{src/environment}
+
+% Import the description of waysome
```
neithernut: C&P

matthiasbeyer: Damn.

neithernut:
Mh... I read on latex2man again.
We should have the following manpage-sections:
- Synopsis
- Description
- Options
- Files
- See Also (refering to the more detailed PDF version or website)
- Author

Those would also be `\section`s in the manpage's TeX file. However, we don't necessarily want them to be sections of the PDF version.
One solution would be to have an `\input` including the sources for all those sections directly into one section (which is followed by sections about the API and some other notes) while splitting them up into multiple sections for the man page.

neithernut:
ACK for squashing.

neithernut:
ACK.

PR #404, opened 2014-11-14
Run a transaction on an event of the same name

neithernut:
This PR modifies the action manager's behaviour by introducign a "shortcut" or convenience functionality. Previously, to trigger a transaction, there has to be a mapping from an event name to said transaction. This PR introduces direct execution of transactions, if no mapping with the name of the event was found.


matthiasbeyer:
ACK.

PR #405, opened 2014-11-16
Add connbuf unblocking convenience macro

neithernut:
This PR enhances the readability of the connector's code by introducing a convenience macro, as suggested by @jonashaag.
Please do not merge right away but let me do the merge.

### DIFF
--- a/src/connection/connbuf.c
+++ b/src/connection/connbuf.c
@@ -110,6 +110,14 @@ ws_connbuf_append(
 }
 
 int
+ws_connbuf_unblock(
+    struct ws_connbuf* self //!< The object
+) {
+    self->blocked = false;
+    return 0;
+}
+
+int
 ws_connbuf_discard(
     struct ws_connbuf* self,
     size_t amount
--- a/src/connection/connbuf.c
+++ b/src/connection/connbuf.c
@@ -104,9 +104,7 @@ ws_connbuf_append(
     }
 
     self->data += amount;
-    self->blocked = false;
-
-    return 0;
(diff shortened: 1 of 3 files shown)


jonashaag:
Btw this doesn't need to be a macro since compilers can optimize this away easily.

The advantage of not using a macro here is that external code using that module doesn't need to be recompiled when the internal implementation of `ws_*` changes.

neithernut:
Yes, it can be optimized _if_ we use LTO.
I'll change it anyways. We can then rewrite `ws_connbuf_append()` to use `ws_connbuf_unblock()`, which _kind of_ makes sense.

matthiasbeyer:
@jonashaag Are you fine with a "Suggested-by:" tag in the commit message?

src/connection/connbuf.c
```diff
@@ -104,8 +104,14 @@ ws_connbuf_append(
     }
 
     self->data += amount;
-    self->blocked = false;
+    return ws_connbuf_unblock(self);
+}
 
+int
+ws_connbuf_unblock(
+    struct ws_connbuf* self //!< The object
```
matthiasbeyer, line 112: You can remove the comment on this line!

matthiasbeyer:
Besides the one annotation: ACK. Smart PR!

Wait for an answer from @jonashaag , though!

jonashaag:
I don't care

neithernut:
In this case, the PR is ready for merge.

matthiasbeyer:
Feel free to do so.

Issue #339, opened 2014-10-29
Framework for transaction testing


matthiasbeyer:
I guess we should write a framework for transaction testing. I don't know whether this can be done with check, but I imagine something like a bunch of transactions in `tests/check/transactions/` or something, where each one gets loaded and parsed and then verified.

I consider this as important. The infrastructure should be capable to do the reverse as well: Having some commands and translating it to JSON (serializing it). Lateron we could test whether a transaction can be parsed to the internal representation, verify it and then translate it back to JSON and verify it against the input in the first place.

PR #406, opened 2014-11-17
Add transaction testing framework

matthiasbeyer:
Please note: This is WIP.

Targets #339.

My approach: I have exactly one testing function which gets called in a loop,
check passes a `int` to it, which iteration is run atm. I use this `int`
as array index for the `inputs` array, which stores a input (which gets
deserialized) and an expeced output (which gets generated from the deserialized
message).

This can _only_ be done with "event" messages, as this is the only type
of message both the client and the "server" have to be able to talk.

---

Thoughts on this:
1. An event from the client to the server is not the same as an event from
   the server to the client. I guess we should redefine the API to look the same,
   shouldn't we? I consider this as important, but _NOT_ before the end of the
   project phase. We have more urgent things to do.
2. By now, this testing stuff has always buffers which are large enough.
   Other tests (with smaller buffers) can be added easily, but I consider this as
   secondary as well. An issue should be opened for this (with prio:low), right?

### DIFF
--- a/tests/check/serialize/json_transaction/test.c
+++ b/tests/check/serialize/json_transaction/test.c
@@ -48,7 +48,59 @@
 
 #include "serialize/deserializer.h"
 #include "serialize/serializer.h"
+#include "serialize/json/deserializer.h"
+#include "serialize/json/serializer.h"
+#include "objects/message/message.h"
+#include "serialize/json/keys.h"
+#include "util/arithmetical.h"
 
+static const struct {
+    const char* in;
+    const char* out;
+} inputs[] = {
+    {
+        .in =   "{"
+                    "\"" TYPE "\": \"" TYPE_EVENT "\","
+                    "\"" EVENT_NAME "\": \"testevent\","
+                    "\"" EVENT_VALUE "\": 1"
+                "}",
+
+        .out =  "{\"event\":{\"context\":1,\"name\":\"testevent\"}}"
+    },
+};
+
+/*
+ *
+ * Test
+ *
+ */
+START_TEST (test_inout) {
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@neithernut , @TheNeikos thoughts?

matthiasbeyer:
I consider this as ready but will add tests if you want me to. Please review.

tests/check/serialize/json_transaction/test.c
```diff
@@ -0,0 +1,164 @@
...
+
+    struct ws_message* message = NULL;
+    const char* msg = inputs[i].in;
+
+    ws_deserialize(d, &message, msg, strlen(msg));
+    ck_assert(message);
+
+    size_t nbuf = strlen(msg);
+    char buf[nbuf];
+    memset(buf, 0, nbuf);
```
manuelmessner: Probably a simple `buf[nbuf - 1] = 0;` would be enough

neithernut:
I'd rather have the framework check against the embedded value instead of the _exact string_.

matthiasbeyer:
This is covered by the de/serializer specific tests. This test "framework" is just for checking whether the whole construct works as expected.

neithernut:
Excactely. So why rely on the excact representaition if you don't need the deserialization in the first place?

matthiasbeyer:
Sorry, but I still don't get what you mean here...

neithernut:
Instead of writing out the reply, we should _only_ check whether the return vlalue _itself_ is correct. Hence, we can simply compare the value which the reply contains instead of serializing it and `strcmp()`ing _that_.

PR #407, opened 2014-11-17
[Refactor] Remove "run" callback

matthiasbeyer:
This PR removes the "run" callback from the `ws_object` type and all
callback tables.

The intention behind the run callback was that we might have objects
which are runnable. Turns out we don't have objects which need to be run.

This rips the code out of the codebase. Please review and/or discuss.


neithernut:
ACK.

Issue #366, opened 2014-11-03
Add global context object for use in transaction.


neithernut:
When a transaction is run, an object should be pushed to the stack prior to running the transaction, which serves as a kind of "global context".
Users may use this e.g. to query other objects to involke methods on.

neithernut:
Blocks #400.

PR #410, opened 2014-11-17
Added global context

TheNeikos:


neithernut:
Tergets #366.

src/action/context.c
```diff
@@ -0,0 +1,114 @@
...
+static int
+func_log(
+    union ws_value_union* stack
+) {
+    int n = 1;
+    struct ws_string* ctx_str = NULL;
+    struct ws_string* str = NULL;
+    int lvl = 0;
+    union ws_value_union cur;
+    while((cur = stack[n]).value.type) {
```
matthiasbeyer: _puke_

Please do this on several lines. It looks ugly as hell!

matthiasbeyer: You can simply use a `for` loop here, and everything looks fine, I guess.

src/action/context.c
```diff
@@ -0,0 +1,114 @@
...
+}
+
+static int
+func_log(
+    union ws_value_union* stack
+) {
+    int n = 1;
+    struct ws_string* ctx_str = NULL;
+    struct ws_string* str = NULL;
+    int lvl = 0;
```
matthiasbeyer, line 84: I guess we should assume `LOG_INFO` as default here!

src/action/context.c
```diff
@@ -0,0 +1,114 @@
...
+                str = ws_value_string_get(&cur.string);
+            }
+            break;
+
+        }
+        n++;
+    }
+
+    if (ctx_str && str && lvl) {
+        struct ws_logger_context log_ctx = { .prefix = ws_string_raw(ctx_str) };
```
matthiasbeyer, line 110: Shall we add a prefix like "[Client: <client id>]" here? Maybe not yet, but I guess it would be nice to have later.

src/action/context.c
TheNeikos, line 84: No, do we have optional values?

src/action/context.c
TheNeikos: It is readable. I could add a comment that the NIL value type is 0 in the enum

src/action/context.c
matthiasbeyer, line 84: What do you mean?

src/action/context.c
matthiasbeyer: I guess this discussion is really opinion-based, so we shouldn't discuss this.

After all, a `for` loop seems to be the right choice to me, as you can `for (cur = stack; cur.value.type; cur++) {` here, couldn't you?

src/action/context.c
TheNeikos, line 84: You need to specify all three arguments for this function to work.

matthiasbeyer, line 84: Ok.

src/action/context.c
```diff
@@ -0,0 +1,115 @@
...
+) {
+    //!< @todo: Make it exit
+    return 0;
+}
+
+static int
+func_log(
+    union ws_value_union* stack
+) {
+    int n = 0;
```
matthiasbeyer, line 81: I guess you don't need that variable anymore, right?

TheNeikos, line 81: I do, and I am using it. I need to know which is the current arguemnt.

matthiasbeyer, line 81: Ah, I overlooked the use. I'm sorry.

matthiasbeyer:
ACK.

neithernut:
ACK.

PR #411, opened 2014-11-17
Reset hotkeys state everytime it's `NULL`

neithernut:
This fixes a SEGFAULT we just encountered.


matthiasbeyer:
Looks good to me. ACK.

Issue #330, opened 2014-10-28
Add push and pop command


neithernut:
We need some commands for pushing and popping.
Pushing would be a special command. Popping should also be implemented as a special command.
The commands would go into the `control` category of commands.

xVin:
I'm trying to do this atm. ;)

PR #414, opened 2014-11-17
Add pop/push command

matthiasbeyer:
Targets #330.

Please review and comment.

### DIFF
--- a/src/command/control.c
+++ b/src/command/control.c
@@ -26,6 +26,7 @@
  */
 
 #include <errno.h>
+#include <string.h>
 
 #include "action/processor.h"
 #include "action/processor_stack.h"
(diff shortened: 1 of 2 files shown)


neithernut:
For pushs and pops, we also want to be able to pass the number of items just like with a "jump" command: By the _number_ of arguments.

matthiasbeyer:
You mean I can pass (for example) three arguments to `pop` and it pops three values from the stack?

I can understand that for the `push` command, where you push all arguments from the command on the stack, but why for the `pop` command?

neithernut:
I think you misunderstand. With the `jump` command, you can write somthing like:

```
{cmd:'jump', attributes : 3}
```

instead of 

```
{cmd:'jump', attributes : {'int' : 3}}
```

With your current implementation, this is not possible with `push` and `pop`.

matthiasbeyer:
Wait a second! The (de)serializer does not support this! We decided to embedd simple values directly into the JSON (simple values = JSON supported values). So your second example wouldn't get parsed.

The complete command-call would look like this:

``` json
{  "jump" : [ 3 ] }
```

We erased the `"jump"` and `"arguments"` keys for simplicity of the parser.

You also need an array for the arguments. I guess we support `arguments : 3` for single arguments, but I'm not sure whether this is implemented properly. Note that the argument gets embedded directly, so the `3` here would result in a `struct ws_value_int` instance internally.

For a complete example [look here](https://github.com/waysome/waysome/blob/develop/tests/check/serialize/json_deserializer/test.c#L251).

neithernut:
Frankly, I don't _care_ what's supported by the deserializrer atm.
To have this functionality in the processor _may_ prove useful afterwards, e.g. if we want to support this.

matthiasbeyer:
So you mean I should be able to specify the number of values to pop directly via

``` json
{ "pop": [ 3 ] }
```

And indirectly via an integer value from the stack by doing

``` json
{ "pop": [ { "pos": -15 } ] }
```

for example? For the push command, this would result in either

``` json
{ "push": [ { "pos": -1 }, { "pos": -5 } ] }
```

For pushing the values on position `-1` and `-5` from the stack directly on the stack again. This sounds sane. For passing the value to push directly from the command, this would look like

``` json
{ "push": [ 1, "hi there", false ] }
```

for pushing an integer, string and boolean, for example.

---

Is this the behaviour I shall implement here?

neithernut:
Not indirectly, but _implicitely_.

src/command/control.c
```diff
@@ -130,3 +131,83 @@ ws_builtin_cmd_store(
...
+            } else { // It is real!
+                v = args->vals[i].arg.val;
+            }
+
+            if (!v) {
+                //!< @todo error?
+                continue;
+            }
+            ws_value_union_init_from_val(&top[-i], v);
+        }
```
neithernut, line 211: This will result in the arguments being present in reverse order.

matthiasbeyer, line 211: Fixed.

neithernut, line 211: Nope, it's _not_. Reversing the loop iteration doesn't help if the problem is your _indexing_ of the two arrays.

matthiasbeyer, line 211: Please review this part again, I'm not sure about the `-1`, but I guess it is right.

neithernut, line 211: The `-1` is one too many.
Why don't you just `top -= n` and use `top[i]` within the loop? would be much clearer that way.

matthiasbeyer, line 211: This is indeed a good suggestion!

neithernut:
ACK for squashing.

neithernut:
ACK.

Issue #425, opened 2014-11-19
Add exec API function


neithernut:
We want some facility for starting programs through transactions.

matthiasbeyer:
I guess this includes some utility for fork & exec, right?

neithernut:
Yep.

PR #428, opened 2014-11-19
Add command for starting an application

matthiasbeyer:
This is WIP and targets #425.

@neithernut please review my exec() utility function, which is the first
step towards a command.

### DIFF
--- a/src/util/exec.h
+++ b/src/util/exec.h
@@ -1,0 +1,52 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
+ * Copyright (C) 2014-2015 Manuel Messner
+ * Copyright (C) 2014-2015 Marcel Müller
+ * Copyright (C) 2014-2015 Matthias Beyer
+ * Copyright (C) 2014-2015 Nadja Sommerfeld
+ *
+ * This file is part of waysome.
+ *
+ * waysome is free software: you can redistribute it and/or modify it under the
+ * terms of the GNU Lesser General Public License as published by the Free
+ * Software Foundation, either version 2.1 of the License, or (at your option)
+ * any later version.
+ *
+ * waysome is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+ * A PARTICULAR PURPOSE. See the GNU Lesser General Public License for more
+ * details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+/**
+ * @addtogroup utils "(internal) utilities"
+ *
+ * @{
+ */
+
+#ifndef __WS_UTIL_EXEC_H__
+#define __WS_UTIL_EXEC_H__
+
+#include "util/attributes.h"
+
+int
+ws_exec(
+    const char* call,
+    char* const argv[]
+)
+__ws_nonnull__(1)
+__ws_noreturn__
+;
+
+#endif // __WS_UTIL_EXEC_H__
+
+/**
+ * @}
+ */
--- a/src/util/exec.h
+++ b/src/util/exec.h
@@ -42,7 +42,6 @@ ws_exec(
     char* const argv[]
 )
 __ws_nonnull__(1)
-__ws_noreturn__
 ;
 
 #endif // __WS_UTIL_EXEC_H__
--- a/src/context.c
+++ b/src/context.c
@@ -49,9 +49,18 @@ func_log(
     union ws_value_union* stack
 );
 
+/**
+ * A way to start a programm
+ */
+static int
+func_exec(
+    union ws_value_union* stack
+);
+
 static const struct ws_object_function functions[] = {
     { .name = "exit", .func = func_exit },
     { .name = "log", .func = func_log },
+    { .name = "exec", .func = func_exec },
     { .name = NULL, .func = NULL }
 };
 
@@ -116,3 +125,12 @@ func_log(
 
     return 1;
 }
+
+static int
+func_exec(
+    union ws_value_union* stack
+) {
+    //!< @todo implement
+    return 1;
+}
+
--- a/src/context.c
+++ b/src/context.c
@@ -25,12 +25,14 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
 #include <ev.h>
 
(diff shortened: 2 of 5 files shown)


src/util/exec.h
```diff
@@ -0,0 +1,52 @@
...
+
+#include "util/attributes.h"
+
+int
+ws_exec(
+    const char* call,
+    char* const argv[]
+)
+__ws_nonnull__(1)
+__ws_noreturn__
```
neithernut: The function _will_ return!

neithernut:
Please add a function to the global context in `src/action/context.c`.

src/util/exec.h
matthiasbeyer: Not always. If the `exec()` call works, it does not return (as stated in the manpage of the call).

matthiasbeyer:
@neithernut Added, please review. I don't know whether this implementation is sufficient.

What I'm doing: Taking N arguments from the stack, which _all_ have to be strings. If there is a non-string argument, I `-EINVAL`. I collect the actual `char*`s from that strings and pass them to the new `ws_exec()` function as arguments.

No `NULL` checking by now.

src/util/exec.h
neithernut: The attribute is intended for functions that **never** return!

matthiasbeyer: TIL.

src/action/context.c
```diff
@@ -116,3 +127,41 @@ func_log(
...
+
+    while (n) {
+        struct ws_string* arg = ws_value_string_get(&stack[n + 1].string);
+        args[n] = ws_string_raw(arg);
+
+        --n;
+    }
+
+    return ws_exec(cmd, (char**)args);
+}
```
neithernut: The elements in `args` must be freed.
Also: add a blank line at EOF.

matthiasbeyer: Seriousely? The elements from `args` come directly out of a `struct ws_string`, which does the `free()` call when it gets cleaned up!

neithernut: `ws_string_raw` allocates memory for the string it returns. You have to free it!

neithernut:
The first two arguments you get are the object and the command name. You have to ignore them.

src/action/context.c
```diff
@@ -116,3 +127,43 @@ func_log(
...
+    enum ws_value_type iter_type;
+
+    // count number of arguments and do type checking meanwhile
+    for (n = 2, iter_type = ws_value_get_type(&stack[n + 1].value); iter; n++) {
+        if (iter_type != WS_VALUE_TYPE_NIL &&
+                iter_type != WS_VALUE_TYPE_STRING) {
+            return -EINVAL;
+        }
+
+        iter = iter_type != WS_VALUE_TYPE_NIL;
```
neithernut: You _really_ want to look for `WS_VALUE_TYPE_NONE`, and **not** for `WS_VALUE_TYPE_NIL`!

src/action/context.c
```diff
@@ -116,3 +127,43 @@ func_log(
...
+    }
+
+    const char* args[n];
+
+    while (n) {
+        struct ws_string* arg = ws_value_string_get(&stack[n + 1].string);
+        args[n] = ws_string_raw(arg);
+
+        --n;
+    }
```
neithernut: As n is _offset_ by `2`, this will _definitely_ not result in what you want.

src/action/context.c
matthiasbeyer: Indeed.

src/action/context.c
matthiasbeyer: Indeed. Was fixed.

neithernut: No, you did **not** fix it, since you still pass `args` to `ws_exec` **without any offset!**

matthiasbeyer: `args` does not _have_ any offset, as it is a simple `char*[]`. There is no _need_ for an offset here. I compile the `ws_string`s from the arguments into a `char*[]`. In the arguments of the command, I have to use the offset, but not in the `char*[]`.

neithernut: Yep, now **please** read the thing you wrote again.

src/action/context.c
```diff
@@ -116,3 +127,43 @@ func_log(
...
+
+    const char* args[n];
+
+    while (n > 2) {
+        struct ws_string* arg = ws_value_string_get(&stack[n + 1].string);
+        args[n - 2] = ws_string_raw(arg);
+
+        --n;
+    }
```
neithernut: Your code is still faulty. You need to decrement `n` _before_ copying.

matthiasbeyer:
I did a complete reimplementation of the `func_exec()` function, as the code became a mess. Please review again.

src/action/context.c
```diff
@@ -116,3 +127,42 @@ func_log(
...
+
+    const char* args[n];
+
+    while (n) {
+        --n;
+        struct ws_string* arg = ws_value_string_get(&stack[n].string);
+        args[n] = ws_string_raw(arg);
+    }
+
+    return ws_exec(cmd, (char**)args);
```
neithernut: The elements of `args` are still not `free()`d.

matthiasbeyer: The elements of `args` are still directly retreived from a `ws_string` which does the cleanup itself.

neithernut: **NO!!!** They are allocated in `ws_string_raw`, for god's sake!

src/action/context.c
```diff
@@ -116,3 +127,51 @@ func_log(
...
+    stack += 2; // We don't care about object itself and command name
+
+    if (ws_value_get_type(&stack->value) != WS_VALUE_TYPE_STRING) {
+        return -EINVAL;
+    }
+
+    struct ws_string* strcmd    = ws_value_string_get(&stack->string);
+    const char* cmd             = ws_string_raw(strcmd);
+
+    stack++; // because of `cmd` is extracted now
```
xVin: "of" can be removed

src/action/context.c
```diff
@@ -116,3 +127,51 @@ func_log(
...
+    enum ws_value_type iter_type;
+
+    // count number of arguments and do type checking meanwhile
+    for (n = 0, iter_type = ws_value_get_type(&stack[n].value); iter; n++) {
+        if (iter_type != WS_VALUE_TYPE_STRING) {
+            return -EINVAL;
+        }
+
+        iter = iter_type != WS_VALUE_TYPE_NONE;
+    }
```
neithernut: `iter_type` will _not_ be refreshed on every loop iteration (which is definitely not what you want).

neithernut:
Is this still WIP?

matthiasbeyer:
Nope

neithernut:
Then ACK for squashing.

Issue #429, opened 2014-11-19
Add API binding for method call


neithernut:
We forgot completely about _that_. I'm on it.

PR #430, opened 2014-11-19
Add command for calling an object's method

neithernut:
This PR adds a command which was previously missing.
Targets #429.


matthiasbeyer:
ACK.

Issue #422, opened 2014-11-19
Add method for querying the surface having focus to global context


neithernut:
This blocks #420.

matthiasbeyer:
Don't we need a global storage for this first?

neithernut:
The API binding is missing. That's the point of this issue.

PR #431, opened 2014-11-19
Add surface global context

TheNeikos:


neithernut:
Targets #422.

neithernut:
Please rebase.

neithernut:
I'm not sure whether the syle fixes shouldn't be a commit on it's own...
But... ACK

src/context.c
```diff
@@ -298,3 +317,25 @@ func_get_surface_under_cursor(
     return 0;
 }
 
+static int
+func_get_ms_focus(
+    union ws_value_union* stack
+) {
+    struct ws_surface* surface = ws_cursor_get()->active_surface;
```
matthiasbeyer, line 324: It is guaranteed that the `ws_cursor_get()` always returns the expected thing?

src/context.c
```diff
@@ -298,3 +317,25 @@ func_get_surface_under_cursor(
...
+    ws_value_object_id_set(&stack[0].object_id, (struct ws_object*) surface);
+
+    return 0;
+}
+
+static int
+func_get_kb_focus(
+    union ws_value_union* stack
+) {
+    struct ws_surface* surface = ws_keyboard_get()->active_surface;
```
matthiasbeyer, line 335: Same here.

src/context.c
```diff
@@ -298,3 +317,25 @@ func_get_surface_under_cursor(
     return 0;
 }
 
+static int
+func_get_ms_focus(
+    union ws_value_union* stack
+) {
+    struct ws_surface* surface = ws_cursor_get()->active_surface;
+
+    ws_value_object_id_set(&stack[0].object_id, (struct ws_object*) surface);
```
matthiasbeyer, line 326: No need to reinit the stack position before?

src/context.c
```diff
@@ -298,3 +317,25 @@ func_get_surface_under_cursor(
...
+    return 0;
+}
+
+static int
+func_get_kb_focus(
+    union ws_value_union* stack
+) {
+    struct ws_surface* surface = ws_keyboard_get()->active_surface;
+
+    ws_value_object_id_set(&stack[0].object_id, (struct ws_object*) surface);
```
matthiasbeyer, line 337: No need to reinit the stack position before?

src/context.c
TheNeikos, line 324: Yes, it is just a getter for an internal struct

src/context.c
TheNeikos, line 326: No as it is already an object id.

matthiasbeyer:
ACK.

PR #433, opened 2014-11-19
Removed unneccessary dependencies from deserializer test CMakeLists.txt

matthiasbeyer:


neithernut:
Please do the same for the serializer.

matthiasbeyer:
It is already done for the serializer tests. This is actually a copy of the serializers CMakeLists.txt. I tested whether I'm able to remove things, but I couldn't.

neithernut:
ACK, but I may come back later.

PR #435, opened 2014-11-19
Fix module dependencies

neithernut:
This PR moves a few things around, breaking circular dependencies, so we have a nice dependency tree.

This is done by:
- [X] Make "command" _not_ depend on "connection"
- [X] Make "command" _not_ depend on "action"
- [X] Make action manager's context settable from the outside on init

If you see any circular dependencies which I didn't yet mention, notify me.


src/command/command.c
```diff
@@ -86,3 +132,53 @@ ws_command_get(
     return NULL;
 }
 
+int
+ws_command_add(
```
matthiasbeyer, line 136: This should be a plural (`ws_commands_add` instead of `ws_command_add`)

neithernut, line 136: The file's prefix is `ws_command`. If anything, I the entire file should be renamed and **all** the prefixes changed.
I _can_ do this, but I'd rather not.

matthiasbeyer, line 136: Oh shit. I wouldn't rename the functions either, but maybe we could use something like `ws_command_add_commands` here? Would be a bit more verbose about what happens...

neithernut, line 136: Nah... I mean: what's going to be added apart from... commands?

neithernut:
This is ready for merge now.
Note that the contents of `context.c` will have to be moved at some point too, but after merging this we will _at least_ have a starting point for integrating all the functions we currently have sitting around in PRs.

---

Ah fuck it, I'll move `context.c` now.

neithernut:
_Now_ I'm ready. Everyone working on the context should be rebased once this is merged.

---

A future PR will make all the modules real static libraries btw.

src/action/commands.h
```diff
@@ -0,0 +1,39 @@
...
+ * along with waysome. If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#ifndef __WS_ACTION_COMMANDS_H__
+#define __WS_ACTION_COMMANDS_H__
+
+#include "command/command.h"
+
+/**
+ * Add comamnds which have impact to the processor
```
matthiasbeyer: Typo

src/action/commands.h
```diff
@@ -0,0 +1,39 @@
...
+ */
```
matthiasbeyer, line 37: Missing `@return` documentation

src/command/command.c
matthiasbeyer, line 136: Good point.

TheNeikos:
ACK

matthiasbeyer:
ACK.

PR #436, opened 2014-11-20
Make modules static libraries

neithernut:
This PR reorganized the build environmane,t which brings many advantages, e.g. easier creation of tests and faster incremental builds.


matthiasbeyer:
Most of this PR seems to be scripted. I did a short review of the most stuff, but I'd like @TheNeikos to do a review as well.

neithernut:
The creation or some commits was scripted. The dependency deduction was performed by hand.

neithernut:
May this merged already?

matthiasbeyer:
The latest commit has a typo in the commit message.

I'd also suggest that you rebase on the latest develop, to be sure there are no fuckups around...

neithernut:
This should be checked by the CI, as we perform a merge just before comilation and testing now, but i'll do it nonetheless.

---

Done.

matthiasbeyer:
The problem is that the build doesn't get triggered when develop updates, so there is still a way to crash the whole thing.

But well... will review now (again).

src/CMakeLists.txt
```diff
@@ -26,44 +26,16 @@ include_directories(
...
-    ${WAYLAND_CURSOR_INCLUDE_DIRS}
-    ${WAYLAND_SERVER_INCLUDE_DIRS}
-    ${XKB_COMMON_INCLUDE_DIRS}
-    ${YAJL_INCLUDE_DIRS}
-    ${UUID_INCLUDE_DIRS}
 )
 
 #
-# Add definitions
+# Set devintions
```
matthiasbeyer: Typo.

matthiasbeyer:
Please fix the one typo, besides: ACK.

src/CMakeLists.txt
neithernut: I wonder how this happened...

matthiasbeyer: Me two!

neithernut:
Note that you have to re-run CMake/clean the cache when transitting to the new setup. I'd suggest blindly rebasing _every_ PR.

matthiasbeyer:
I rebased everything I have.

Issue #444, opened 2014-11-21
Extend file section of user documentation


neithernut:
This would include
- list of locations where waysome looks for a config file to the user documentation
- sockets opened by waysome
- cursor image

PR #448, opened 2014-11-22
Add content to "files" section of user documentation

matthiasbeyer:
Targets #444

Please review.


neithernut:
Mh... maybe we should use inline listings for the files, but I'm not quite sure how this interacts with `latex2man`...

matthiasbeyer:
What do you mean by "inline listings"? Putting source code in the (user-)documentation?

neithernut:
I mean using `lstinline` for paths, which e.g. causes usage of another font and _maybe_ also colours env-vars used in the path.

matthiasbeyer:
Just discovered that there is `\path` around.

matthiasbeyer:
`\path` does not variables. I opened a question on texoverflow how to do it inside a `\path{}`, but I don't think it is possible. If it is not possible, I will write a command which results in a simple `bash`-listing environment.

I'm now setting this on hold until I get a sufficient answer on texoverflow, but I will not wait that long.

neithernut:
`latex2man` actually sets a "variable" with which one can check whether a manpage is created or not. Using this, we can create a command which does "the right thing" for both manpages and PDFs.

matthiasbeyer:
Can you tell me how to build the man page? My `.1` file only contains the note that it is generated, but no content by now... maybe I've gotten something wrong,... don't know.

neithernut:
I think the manpage.tex is missing some elements. See http://ctan.space-pro.be/tex-archive/support/latex2man/latex2man.html. (maybe open an issue for that)

---

I'll open an issue.

matthiasbeyer:
`latex2man` actually does not understand `\input`. I suggest dropping it, as it is not of any benefit for us anymore.

neithernut:
So, you suggest _not_ to have a man-page?

matthiasbeyer:
Nope. I suggest to dropping `latex2man`.

neithernut:
So you want to hand-craft a manpage?

matthiasbeyer:
Not really.

My point is that using latex2man gives us no benefit over writing it by hand, except for writing it in latex. So, the initial point of writing parts of the documentation once and have them both in the pdf and the manpage just vanished.

Although I see no alternatives between latex2man and writing one by hand.

---

What I'd like to suggest: Put the whole manpage stuff on the back burner. We can write a manpage when we are in v1.0, but it is really not worth the effort right now.

neithernut:
k, so you'll write one by hand. (btw: I did this once)

matthiasbeyer:
I guess this shouldn't be discussed here anyways, as this PR targets on another topic. So,... what about the patch now?

neithernut:
Was marked as WIP and on hold, so I didn't bother reviewing it.
Now that I did: ACK.

PR #455, opened 2014-11-22
Silence pdflatex

neithernut:
`pdflatex` spams a lot of output onto stdout. We don't need that output since it also produces a transcript.


TheNeikos:
ACK if pdflatex outputs on stdout

neithernut:
It does.

PR #460, opened 2014-11-22
Fix cmd_func_set_pos

matthiasbeyer:
Actually, we're calling the command on an object. This object is the
shell_surface instance we want to alter here, not the first argument
passed to the command, right? So we should use this object to do our
transformation on, not something else.

@neithernut Please review.

### DIFF
--- a/src/compositor/wayland/shell_surface.c
+++ b/src/compositor/wayland/shell_surface.c
@@ -422,19 +422,23 @@ static int
 cmd_func_set_pos(
     union ws_value_union* stack
 ) {
-    stack += 2; // Ignore the object and the command name
-
-    if (ws_value_get_type(&stack[0].value) != WS_VALUE_TYPE_OBJECT_ID ||
-            ws_value_get_type(&stack[1].value) != WS_VALUE_TYPE_INT ||
-            ws_value_get_type(&stack[2].value) != WS_VALUE_TYPE_INT ||
-            ws_value_get_type(&stack[3].value) != WS_VALUE_TYPE_NIL) {
+    if (ws_value_get_type(&stack[0].value) != WS_VALUE_TYPE_OBJECT_ID) {
         return -EINVAL;
     }
 
     struct wl_resource* r = (struct wl_resource*)
                             ws_value_object_id_get(&stack[0].object_id);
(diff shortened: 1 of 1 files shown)


TheNeikos:
ACK

manuelmessner:
ACK

neithernut:
The fist two arguments _any_ command callback set on an object see's is the object itself, the second one is the method name. So this code will fail horribly for any case where the code is actually called.

matthiasbeyer:
What "code" are you talking about here?

The first argument is the object, of course. That's the whole point of the PR, as we want to call an method on this object, we need to check it. Afterwards we check the arguments of the command. So I don't get where my error is...?

src/compositor/wayland/shell_surface.c
```diff
@@ -422,19 +422,23 @@ static int
...
+        return -EINVAL;
+    }
+
+    struct wl_resource* r = (struct wl_resource*)
+                            ws_value_object_id_get(&stack[0].object_id);
+
     stack += 2; // Ignore the object and the command name
 
-    if (ws_value_get_type(&stack[0].value) != WS_VALUE_TYPE_OBJECT_ID ||
+    if (ws_value_get_type(&stack[0].value) != WS_VALUE_TYPE_INT ||
```
neithernut, line 434: stack[0] will _always_ be an object id!

src/compositor/wayland/shell_surface.c
```diff
@@ -422,19 +422,23 @@ static int
...
             ws_value_get_type(&stack[1].value) != WS_VALUE_TYPE_INT ||
```
neithernut, line 435: stack[1] will _always_ be a string!

neithernut:
Your entire argument list is offset!

src/compositor/wayland/shell_surface.c
matthiasbeyer, line 434: Not after `stack += 2`.

src/compositor/wayland/shell_surface.c
matthiasbeyer, line 435: Not after `stack += 2`.

neithernut:
Ah, _now_ I get it. Ok, than I can ACK this as well.

Issue #400, opened 2014-11-12
Add commands for adding hotkey events to the context object


neithernut:
There must be functions for adding hotkey events. While the backend versions of those functions exist already, they still have to be exported through the API.

xVin:
I'd like to do this. Could start tomorrow.

neithernut:
Sure, you have to wait for the backend to be merged, though.

neithernut:
Deps are cleared now.

matthiasbeyer:
> There must be functions for adding hotkey events. While the backend versions of those functions exist already, they still have to be exported through the API.

Which functions have to be exported?

Also: Can you elaborate on how the API calls must look like (meta-language is much enough here) and how they behave?

neithernut:
- `ws_hotkey_add` taking a string, followed by ints as arguments
- `ws_hotkey_remove` taking ints as arguments

PR #461, opened 2014-11-22
Add add_hotkey_event() to context.c

xVin:
targets #400

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -27,10 +27,14 @@
 
 #include <errno.h>
 #include <ev.h>
+#include <stdint.h>
 
 #include "context.h"
+#include "command/util.h"
+#include "input/hotkeys.h"
 #include "objects/object.h"
 #include "objects/string.h"
+#include "values/string.h"
 #include "values/union.h"
 #include "util/exec.h"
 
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Merge conflicts would occur. Please rebase.

xVin:
Trying to... please see irc query.

xVin:
Pls review

src/context.c
```diff
@@ -176,3 +180,77 @@ func_exec(
...
+    struct ws_string* str;
+
+    stack += 2; //Ignoring object and command name
+
+    if (ws_value_get_type(&stack->value) != WS_VALUE_TYPE_STRING) {
+        return -EINVAL;
+    }
+
+    vstr = &stack->string;
+    str = ws_value_string_get(vstr);
```
matthiasbeyer: You can merge this and the line above.

src/context.c
```diff
@@ -176,3 +180,77 @@ func_exec(
...
+    }
+
+    size_t i = 0;
+    uint16_t arr[len];
+    ITERATE_ARGS_TYPE(it, stack, v, int) {
+        arr[i] = (uint16_t) v;
+        ++i;
+    }
+
+    return ws_hotkey_add(str, arr, len);
```
matthiasbeyer: Return value should be on stack as well. (maybe as boolean type...)

src/context.c
```diff
@@ -176,3 +180,77 @@ func_exec(
...
+    }
+
+    size_t i = 0;
+    uint16_t arr[len];
+    ITERATE_ARGS_TYPE(it, stack, v, int) {
+        arr[i] = (uint16_t) v;
+        ++i;
+    }
+
+    return ws_hotkey_remove(arr, len);
```
matthiasbeyer: Return value should be on stack, maybe as boolean.

manuelmessner:
ACK

neithernut:
Nice, but you might be better off pulling out the code where you assemble the array into a static helper function.

xVin:
Hm, but that would be pulling out only 3 lines of code that are used in 2 methods, is that really worth it or am I overlooking sth?

neithernut:
Mh.. I thought you can also pull out hte _counting_ of items into the same method, but I realized the advantages are nullified by the fact that you'd have to allocate the memory on the heap then...
So,... ACK.

Issue #457, opened 2014-11-22
Add method for setting a surface's width and height


neithernut:
To perform resizes. This also includes exporting them via the API.

PR #462, opened 2014-11-22
Add width/height setter methods for ws_shell_surface

matthiasbeyer:
Targets #457.

I'm not sure about the convenience method, which allows setting of the
width and height in one step, but I think it is basically a good idea.

### DIFF
--- a/src/compositor/wayland/abstract_shell_surface.c
+++ b/src/compositor/wayland/abstract_shell_surface.c
@@ -76,6 +76,21 @@ cmd_func_set_height(
     union ws_value_union* stack // The stack to use
 );
 
+/**
+ * Callback command function for setting surface width and height
+ *
+ * @memberof ws_abstract_shell_surface
+ *
+ * Takes two parameters:
+ *  1) The object id of the surface
+ *  2) The width to set
+ *  2) The height to set
+ */
+static int
+cmd_func_set_width_and_height(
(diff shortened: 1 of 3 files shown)


matthiasbeyer:
@TheNeikos please review here!

neithernut:
Given the wayland interface, it makes more sense updating width and height at the same time.
Please have a look at `wl_surface`, `wl_shell_surface` and the like. E.g. `wl_shell_surface_send_configure()` is of interest.

matthiasbeyer:
Hm. I think it's suiteable to provide some convenience here and export setters for width and height only as well... don't you think so?
@TheNeikos what about you?

neithernut:
All the methods need to be relocated into either `ws_abstract_shell_surface` or `ws_surface`.

matthiasbeyer:
I'd like to wait for #539 , so I can avoid heavy rebasing here.

matthiasbeyer:
I did a complete rewrite of the branch. Please review.

neithernut:
ACK so far, but could you try to implement the `todo`s you added in this PR? (e.g. using the set-width-and-height-method for setting width and height, setting the other value to the current one).

matthiasbeyer:
Of course I will do this!

neithernut:
ACK for squashing.

matthiasbeyer:
done.

Issue #451, opened 2014-11-22
Not all `ws_object` type methods are thread safe


matthiasbeyer:
Mainly speaking of `ws_object_instance_of()` here, but I guess others aren't, too. I will investigate.

neithernut:
This special method doesn't need to be thread save as an arguments type doesn't change during runtime! This also holds true for a few other methods.

matthiasbeyer:
I will investigate on the types anyways and improve where necessary.

matthiasbeyer:
Just found: does the `ws_keyboard_send_keymap()` need locking on the passed `ws_keyboard` object? @TheNeikos ? 

---

Same for `..._send_enter()` and the like...

matthiasbeyer:
Closed as the appropriate PR got merged.

PR #465, opened 2014-11-22
Rework object locking

matthiasbeyer:
This PR reworks the object locking.

It
- Adds note why locking is not required in some cases
- Adds locking where it is missing
- Adds ref counting where it is missing

This is **WIP**, the following modules are on my todo-list:
- [X] `src/objects/`
- [ ] `src/compositor/`
- [x] `src/action`
- [x] `src/serialize`
- [x] `src/values`

Feel free to propose more, participate in this PR (just notify me which stuff
you want to do) or the like!

### DIFF
--- a/src/objects/object.c
+++ b/src/objects/object.c
@@ -299,11 +299,11 @@ ws_object_lock_try_read(
     return -pthread_rwlock_tryrdlock(&self->rw_lock);
 }
 
-bool
+int
 ws_object_lock_try_write(
     struct ws_object* self
 ) {
-    return 0 == pthread_rwlock_trywrlock(&self->rw_lock);
+    return -pthread_rwlock_trywrlock(&self->rw_lock);
 }
 
 bool
--- a/src/objects/object.c
+++ b/src/objects/object.c
@@ -292,11 +292,11 @@ ws_object_lock_write(
     return 0 == pthread_rwlock_wrlock(&self->rw_lock);
 }
 
-bool
+int
 ws_object_lock_try_read(
     struct ws_object* self
 ) {
-    return 0 == pthread_rwlock_tryrdlock(&self->rw_lock);
+    return -pthread_rwlock_tryrdlock(&self->rw_lock);
 }
 
 bool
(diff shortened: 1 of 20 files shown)


matthiasbeyer:
Targets #451

neithernut:
You forgot to rebase this one...

matthiasbeyer:
I'm not sure whether I am able to review the compositor code that easily. I would recommend we merge this now and review the compositor code in another step.

src/objects/named.c
```diff
@@ -88,6 +88,8 @@ ws_named_init(
         self->val = getref(val);
     }
 
+    getref(self);
```
neithernut: Doesn't make sense.

src/objects/object.c
```diff
@@ -506,21 +506,38 @@ ws_object_cmp(
...
     }
 
-    return type->cmp_callback(o1, o2);
+    res = type->cmp_callback(o1, o2);
+
+out:
+    ws_object_unlock((struct ws_object*) o2); // unlock, remove `const`
+    ws_object_unlock((struct ws_object*) o1); // unlock, remove `const`
+    return res;
 }
```
neithernut, line 553: This will cause deadlocks.

matthiasbeyer, line 553: How to resolve it?

neithernut, line 553: Loop with try-locks.

matthiasbeyer, line 553: This could be done on the inner (the second) lock only, right?

neithernut, line 553: Not quite, you also have to release the outer lock, if the inner one fails.

neithernut, line 553: You should still loop instead of returning `-EAGAIN`.

matthiasbeyer, line 553: This was done in the last push, the diff is outdated.

neithernut:
ACK for squashing.

matthiasbeyer:
Done with squashing.

neithernut:
ACK.

Issue #259, opened 2014-10-21
Add connection accept()ing code


neithernut:
We need to create a socket plugins can connect to. We also have to `accept()` connections from it.

matthiasbeyer:
Is there anything done yet?

manuelmessner:
> Is there anything done yet?

Yes there was nearly everything done when my hard disk crashed.
My system is reinstalled and configured right now. So i'll rewrite it.
It should be finished soon.

matthiasbeyer:
I hope you get the point of `git push` now.

manuelmessner:
yeah -.-

matthiasbeyer:
:angry:

matthiasbeyer:
@manuelmessner There is still no PR around, are you actually doing something?

manuelmessner:
You were sitting next to me when I was working on it today...

neithernut:
Point is: when there's a PR, we can _assess_ the progress.

manuelmessner:
I'll upload my PR when i have a working version or at least when it is nearly working, otherways the history will be a mess.

neithernut:
You can still rebase a branch/squash commits after submitting a PR (as long as it isn't merged). The only thing you _can't_ do is rename the branch.

matthiasbeyer:
@manuelmessner and me are currently on this.

@neithernut Please verify that the following is right:
- The accept()ing code creates the `ws_connection_manager\* instance for the connection
- ~~The accept()ing code creates _ws_connbuf_ instances for reading/writing~~

How does the code interact with the rest of the system, mainly speaking of de/serialization here? What to do with the `ws_connection_manager` after it got instantiated?

neithernut:
The accepting code creates _processors_. Managers are __always singletons!_

Issue #184, opened 2014-10-12
Adding a unix socket to read all the framebuffers as a big png


TheNeikos:
This is a feature request for down the line.

Basically we would just create a unix socket at some specified place and if it is read one can get the screen contents in PNG format (Good support, no additional dependencies and lossless).

libev supports events on sockets, so this shouldn't be too hard to implement.

matthiasbeyer:
:+1: I like this.

neithernut:
The idea is intriguing. Could be made an optional feature (possibility to enable it at runtime).
For the socket's path, some thing like "/tmp/waysome-screen" miight be hardcoded for now.

matthiasbeyer:
Talking about security here: There should be an option to restrict access to this on the waysome side, something like "3 times per minute, but not more" or something. My concerns are also about load. If someone wants to "stream" the screen over this, this would cause waysome to get stuck at some point, I guess.

TheNeikos:
That is definitely a possiblity, although once we have double buffering you would just get the front buffer (the one being drawn), so that way there is no lock on it. (And thus no delay)

Rate limiting shouldn't be too hard to implement, I don't see the usefulness though (You control your machine after all).

What do you mean restricting access on the waysome side?
The socket would only be readable by the user by default imo, so any security problems are on the part of the OS.

neithernut:
Load balancing is a trivial matter.

TheNeikos:
@neithernut How so?

matthiasbeyer:
I meant that the software itself does the "no, I don't write this buffer for the next N seconds" or something, and not some plugin. But okay, I'm not the person to ask these questions!

neithernut:
Most naturally, you'd put the PNG creation routine in a central place, so you can serve multiple clients requesting a screenshot without any extra effort. Updates of the buffer by redraws may mark the buffer as invalid. The next client requesting sees that the PNG is stale and commences a serialisation of the current buffer, every other client requestign an update of the same buffer will get the same PNG, without generating more effort. This way, you generate PNGs once an update, at most.

Optionally, you can introduce a timeout which marks the PNG as stale after a specified amount of time has passed (given that the buffer changed in the meantime). This would generate the behaviour requested by @matthiasbeyer.

neithernut:
Note: you can actually `cat` a socket, so we don't even need some "client" to fetch screenshots, if we deliver _just_ a PNG.

TheNeikos:
Yup! That was the idea

PR #466, opened 2014-11-22
Add unix socket

manuelmessner:
Facing #259

### DIFF
--- a/src/util/socket.c
+++ b/src/util/socket.c
@@ -42,6 +42,19 @@
 
 static struct ws_logger_context log_ctx = { .prefix = "[Sockets Utils] " };
 
+static void
+socket_read_cb(
+    struct ev_loop* loop,
+    struct ev_io* io,
+    int revents
+) {
+    //!< @todo I'll add this when ws_(de)serializer_new is implemented
+    if (revents & EV_ERROR) {
+        ws_log(&log_ctx, LOG_WARNING, "Libev error in read callback!");
+        return;
+    }
+}
+
 int
 ws_socket_create(
     char const* name
--- a/src/util/socket.c
+++ b/src/util/socket.c
@@ -104,6 +104,28 @@ socket_accept_cb(
 }
 
 int
+ws_socket_init(
+    struct ws_socket* s
+) {
+    struct ev_loop* loop = ev_default_loop(EVFLAG_AUTO);
+    if (!loop) {
+        ws_log(&log_ctx, LOG_WARNING, "Could not initialize ev loop!");
+        free(s);
+        return -1;
+    }
+
+    s->fd = ws_socket_create(UNIX_PATH);
+    if (s->fd < 0) {
+        return -1;
+    }
+
+    ev_io_init(&s->io, socket_accept_cb, s->fd, EV_READ);
+    ev_io_start(loop, &s->io);
+
+    return 0;
+}
+
+int
 ws_socket_create(
     char const* name
 ) {
--- a/src/util/socket.c
+++ b/src/util/socket.c
@@ -25,6 +25,7 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
--- a/src/util/socket.c
+++ b/src/util/socket.c
@@ -37,6 +37,7 @@
 #include "util/socket.h"
 
 #define XDG_RUNTIME_DIR "XDG_RUNTIME_DIR"
+#define UNIX_PATH "waysome.sock"
 #define UNIX_PATH_MAX 108
 
 static struct ws_logger_context log_ctx = { .prefix = "[Sockets Utils] " };
--- a/src/util/socket.c
+++ b/src/util/socket.c
@@ -55,6 +55,54 @@ socket_read_cb(
     }
 }
 
+static void
+socket_accept_cb(
+    struct ev_loop* loop,
+    struct ev_io* io,
+    int revents
+) {
+    struct ws_socket* s = (struct ws_socket*) io;
+    struct ws_socket_client* c = NULL;
+    struct sockaddr_un addr;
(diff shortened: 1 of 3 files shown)


src/util/socket.h
```diff
@@ -25,9 +25,61 @@
...
+
+#include "connection/processor.h"
+
+/**
+ * Impementation of the socket types.
+ */
+
+struct ws_socket {
+    struct ev_io io;    //!< ev_io object for libev
+    int fd;             //!< file descriptor of the created socket
```
matthiasbeyer: Documentation for the type missing.

Also the visibility of its members are missing.

src/util/socket.h
```diff
@@ -25,9 +25,61 @@
...
+};
+
+struct ws_socket_client {
+    struct ev_io io;                    //!< ev_io object for libev
+    struct ws_connection_manager* cm;   //!< reference to a con_man object
+    int fd;                             //!< fd of the accepted client
```
matthiasbeyer: Documentation for the type missing.

Also the visibility of its members are missing.

src/util/socket.h
```diff
@@ -25,9 +25,61 @@
...
+ * Impementation of the socket types.
+ */
+
+struct ws_socket {
+    struct ev_io io;    //!< ev_io object for libev
+    int fd;             //!< file descriptor of the created socket
+};
+
+struct ws_socket_client {
+    struct ev_io io;                    //!< ev_io object for libev
```
matthiasbeyer: Would it make sense to inherit from `ws_socket` or does the type differ from the `ws_socket` type semantically?

neithernut:
Needs some work.

matthiasbeyer:
Would continue on this on monday or something... when we are reunited...

src/util/socket.c
```diff
@@ -36,10 +37,112 @@
...
+    memset(&addr, 0, len);
+    c->fd = accept(s->fd, (struct sockaddr*) &addr, &len);
+    if (c->fd < 0) {
+        ws_log(&log_ctx, LOG_ERR, "Could not accept client: %d", errno);
+        free(c);
+        return;
+    }
+
+    /**
+     * @todo: I'll comment the lines in when the given functions are implemented
```
matthiasbeyer, line 89: I guess you don't need to do something like this here. I guess you need to create in and out buffers and pass them to another part of the system, I think @neithernut can explain what's to do here.

matthiasbeyer, line 89: The problem is that you don't know whether the send stuff is JSON or something else. And you don't need to know. You only need to pass the data to the next system, which does the parsing.

neithernut, line 89: **All** you need to do is call `ws_connection_processor_new()` for now. You can safely ignore the return value. We _may_ add a connection _manager_ later and replace direct invocations of `ws_connection_processor_new()`. You can get instances of a serializer and deserializer from the "serialize" subsystem.

I'm _quite_ sure that I told you this more than once, @matthiasbeyer!

matthiasbeyer, line 89: I'm not familiar with the socket stuff, so no, you never told me this before... I guess.

As we only support JSON serialization by now, I assume @manuelmessner shall create a json de/serializer here and we add a more abstract "get a de/serializer"-call for this later on, right?

neithernut, line 89: Yep.

src/util/socket.c
```diff
@@ -36,10 +37,112 @@
...
+) {
+    //!< @todo I'll add this when ws_(de)serializer_new is implemented
+    if (revents & EV_ERROR) {
+        ws_log(&log_ctx, LOG_WARNING, "Libev error in read callback!");
+        return;
+    }
+}
+
+static void
+socket_accept_cb(
```
matthiasbeyer, line 59: @neithernut What do you think: Can the accept callback ripped out here, so we can pass it to the socket from the connection manager? Would be a bit cleaner architecture, I guess.

src/util/socket.c
```diff
@@ -36,10 +37,112 @@
 #include "util/socket.h"
 
 #define XDG_RUNTIME_DIR "XDG_RUNTIME_DIR"
+#define UNIX_PATH "waysome.sock"
 #define UNIX_PATH_MAX 108
 
 static struct ws_logger_context log_ctx = { .prefix = "[Sockets Utils] " };
 
+static void
+socket_read_cb(
```
matthiasbeyer, line 46: @neithernut same as for the accept callback: Shall we pass it to the socket utility?

matthiasbeyer:
Maybe we should also change the `ws_connector` to contain a `ws_socket` but not a file descriptor directly, shouldn't we?

---

Maybe these comments speed things up here a bit...

src/util/socket.c
neithernut, line 59: Ripping out the callback itself doesn't make any sense.
However, it would make things "nicer" if this callback would invoke a callback which in turn spawns the connection. Such a callback would get an fd and return either an `int` (success/failure) or `void`.
This way, the utility can also be used for the screenshot thingy proposed in #184.

src/util/socket.c
neithernut, line 46: Actually, the socket code should do _nothing_ but accepting connections. Reading from them is done by the connection processor.

matthiasbeyer, line 46: So this part should be removed.

neithernut:
I think you **greatly** misunderstand how things work here:
- A `ws_connector` is **one** connection
- A socket is a **source** of connections

If anything, a `ws_socket` should be used in the connection _manager_ (which doesn't exist... yet).

src/util/socket.c
neithernut, line 46: Yep.

src/util/socket.c
matthiasbeyer, line 59: Good idea.

matthiasbeyer:
Whop, yes, I misunderstood that part. Okay... now I also understand why a accept()ing thing should result in `ws_connector` instances...

Issue #467, opened 2014-11-22
Remove `@todo`s code


matthiasbeyer:
This Issue is to track progress of the removals of the `@todo` s in all files.
I will not open an issue for each file, which would be overkill (given the list
of files which contain `@todo`s).

We have 129 `@todo`s in the following files:
- [x] `src/action/manager.c`
- [ ] `src/compositor/buffer/buffer.c`
- [x] `src/compositor/buffer/egl.c`
- [X] `src/compositor/buffer/image.c`
- [ ] `src/compositor/cursor.c`
- [ ] `src/compositor/keyboard.c`
- [ ] `src/compositor/module.c`
- [ ] `src/compositor/monitor.c`
- [X] `src/compositor/wayland/client.c`
- [ ] `src/compositor/wayland/compositor.c`
- [ ] `src/compositor/wayland/keyboard.c`
- [ ] `src/compositor/wayland/pointer.c`
- [ ] `src/compositor/wayland/region.c`
- [ ] `src/compositor/wayland/seat.c`
- [ ] `src/compositor/wayland/seat.h`
- [ ] `src/compositor/wayland/shell_surface.c`
- [ ] `src/compositor/wayland/surface.c`
- [ ] `src/compositor/wayland/xdg_shell.c`
- [ ] `src/compositor/wayland/xdg_surface.c`
- [x] `src/connection/processor.c`
- [ ] `src/input/input_device.c`
- [X] `src/main.c`
- [x] `src/objects/object.c`
- [x] `src/objects/object.h`
- [ ] `src/objects/string.c`
- [X] `src/serialize/json/deserializer.c`
- [ ] `src/serialize/json/deserializer_callbacks.c`
- [x] `src/serialize/json/serializer.c`
- [ ] `src/serialize/json/serializer_state.c`
- [x] `src/values/union.c`
- [x] `src/values/value_type.c`

(Based on commit a243086cd1437dd677820aa88f641a3ab36a193a).

---

To retreive a list files where `@todo`s are located:

``` bash
ack "@todo" -l src/
```

To retreive the number of `@todo`s in the code:

``` bash
ack "@todo" src | wc -l
```

---

Edit: Added `main.c` and sorted list alphabetically.

manuelmessner:
src/main.c is missing right now

matthiasbeyer:
[...]

PR #469, opened 2014-11-22
Enable initialization from name for ws_value_object_id type

matthiasbeyer:
Targets #467.

This is the only `@todo` in the `src/values/value_type.c` file.


manuelmessner:
ACK

neithernut:
ACK.

PR #470, opened 2014-11-22
Add support for WS_VALUE_TYPE_NAMED in ws_value_union_init_from_val()

matthiasbeyer:
Targets #467 (`src/values/union.c`) partially.


manuelmessner:
ACK

neithernut:
What about `object_id` and `set` values? You circumvent refcounting for them.

matthiasbeyer:
Sorry, but I don't get your latest comment.

neithernut:
Sorry, I was rather confusing.
Copying the pointer means you end up with two references to the same value. Now what happens when one of the value gets deinitialized?
I'd rather pleadge for using a value union in named value type and using `ws_value_union_init_from_val()` recursively here.

neithernut:
This is still not acceptable. I suggest rewriting `ws_value_named` here, to hold a value unioon instead of a pointer to a value.

matthiasbeyer:
Pushed for backup, not for review.

matthiasbeyer:
I guess this can be reviewed now...

neithernut:
If we do #511, we can also close this unmerged.

matthiasbeyer:
Of course.

PR #472, opened 2014-11-22
Add real logging call inside main signal handler

matthiasbeyer:
Targets #467

Removes all `@todo`s for `src/main.c`.


manuelmessner:
ACK

neithernut:
ACK.

PR #474, opened 2014-11-22
Removed comments around block in object.c

manuelmessner:
@neithernut 

Pls review.
Is this just a forgotten `@todo`?


neithernut:
Both `ws_string_set_from_raw` and `ws_value_string_set` might fail. You need to account for that.

neithernut:
Good enough. ACK.

PR #486, opened 2014-11-23
Fix style in connector module

matthiasbeyer:
This PR cleanes the style in the connector module (the branchname may not be
appropriate here..., please don't mind!).

This PR also introduces a logic-fix, so it is not _entirely_ about style.

@neithernut Please review carefully (the logic-fix), I am (/ we are)
not entirely sure about it,...

### DIFF
--- a/src/connection/processor.c
+++ b/src/connection/processor.c
@@ -288,7 +288,6 @@ connection_manager_dispatch(
     }
 
 error_handling:
-    //!< error handling
     switch(-res) {
     case -EAGAIN:
     case -EINTR:
--- a/src/connection/processor.c
+++ b/src/connection/processor.c
@@ -368,7 +368,7 @@ connection_manager_flush_msg(
 
         // flush
         res = ws_connector_flush(&proc->conn);
-    } while (res > 0);
+    } while (res == 0);
     return res;
 }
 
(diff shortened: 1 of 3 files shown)


src/connection/processor.c
```diff
@@ -369,7 +368,7 @@ connection_manager_flush_msg(
 
         // flush
         res = ws_connector_flush(&proc->conn);
-    } while (res > 0);
+    } while (res == 0);
```
manuelmessner, line 371: I would like to hear @neithernut 's opinion about that as well.

neithernut, line 371: The initial intent was to try flushing stuff as long as writing to the fd doesn't block. However, this was overlooked when working on the `ws_connector_flush()` method.
As it _should_ return `-EAGAIN` if no more data can be written atm, this _should_ be a legetimate fix.

matthiasbeyer, line 371: Yay, we did something right! :smile:

xVin, line 371: I'm going to write a fix for the flush method.

neithernut:
ACK.

matthiasbeyer:
@neithernut Feel free to merge after you ACKed (also for all other PRs)!

src/connection/processor.c
neithernut, line 371: This isn't neccessary, as it does exactly what it should now. What I meant when I said that this was overlooked was that I overlook this when I was looking at EOF issues.

Issue #478, opened 2014-11-23
Add macro "streq"


matthiasbeyer:
For simple `0 == strcmp(...)`

Also a `strneq()` macro would be nice.

Refactoring of codebase for this macro would be included in this task. Code should go into `util/string.h` or something like that.

PR #487, opened 2014-11-23
Added streq and strneq macro

manuelmessner:
Targets #478

### DIFF
--- a/src/util/string.h
+++ b/src/util/string.h
@@ -45,6 +45,14 @@
 #define ws_streq(str1, str2) \
     (0 == strcmp((str1), (str2)))
 
+/**
+ * Check if the first `n` bytes of `str1` and `str2` are equal
+ *
+ * @return true if the bytes are equal, else false
+ */
+#define ws_strneq(str1, str2, n) \
+    (0 == strncmp((str1), (str2), (n)))
+
 
 #endif // __WS_UTIL_STRING_H__
 
--- a/src/util/string.h
+++ b/src/util/string.h
@@ -35,6 +35,16 @@
 #define __WS_UTIL_STRING_H__
 
 
+#include <string.h>
+
+/**
+ * Check if two strings are equal
+ *
+ * @return true if the two strings are equal, else false
+ */
+#define ws_streq(str1, str2) \
+    (0 == strcmp((str1), (str2)))
+
 
 #endif // __WS_UTIL_STRING_H__
 
--- a/src/util/string.h
+++ b/src/util/string.h
@@ -1,0 +1,44 @@
+/*
+ * waysome - wayland based window manager
+ *
+ * Copyright in alphabetical order:
+ *
+ * Copyright (C) 2014-2015 Julian Ganz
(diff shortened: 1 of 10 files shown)


matthiasbeyer:
Will you continue here and replace code which can be replaced by this?

neithernut:
I'm unhappy with the choice of name for `srtneq` because it looks like "string _not_ equal".

matthiasbeyer:
We discussed this... I guess it's more consistent with `strncmp` and the like... but I understand your point. Suggestions?

manuelmessner:
@matthiasbeyer 

Yes I'm doing this right now

@neithernut 

I'll replace only the streq() as long we didn't got a name for strneq

neithernut:
`streqn`? Mh... sounds weird.

matthiasbeyer:
Btw I would prefer having it with a `ws_` prefix... we already had a problem with a missing prefix, no need to fall into this trap again...

manuelmessner:
@matthiasbeyer 
Changed. I push it later.

@neithernut
what about this?
nstreq => not_string_equal
strneq => string_equal_num

neithernut:
Mh... `strneq` might be the only feasible option...

matthiasbeyer:
Seems so.

@manuelmessner I wouldn't include `nstreq`, as there is no advantage over doing `!streq(...)` ...

manuelmessner:
Ok can I implement that?
`ws_streq` and `ws_strneq`?

matthiasbeyer:
ACK for the first, ACK for the second if @neithernut ACKs it.

neithernut:
Yep, ACK from me, too.

src/util/string.h
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @{
+ */
+
+#ifndef __WS_UTIL_STRING_H__
+#define __WS_UTIL_STRING_H__
+
+
+/**
+ * Compare two strings on equality
```
xVin: "compare sth on" isn't correct English as far as I know. Maybe just write "Check if to strings are equal" if you want to keep it short?

src/util/string.h
```diff
@@ -0,0 +1,57 @@
...
+ *
+ * @return 1 in case of equality, else 0
+ */
+#define streq(x, y) (0 == strcmp((x), (y)))
+
+
+/**
+ * Compare the first n bytes of two strings on equality
```
xVin: see above.

src/util/string.h
matthiasbeyer: "if **to** strings are equal"? I blame you for this!

manuelmessner: Haha :D

@xVin Ok I'll change that.

xVin: lol :D
that's karma

manuelmessner:
ws_strneq is following in a few minutes

manuelmessner:
Pls review

neithernut:
ACK.

Issue #438, opened 2014-11-21
Add method for setting the background of a monitor


neithernut:
This would be a method of `ws_monitor`. The method would take a string as it's only argument, which will be the path of the image to load.
Note that a reference to such an image should be added.

matthiasbeyer:
This is actually not that simple, as the monitor code currently seems to blit the duck manually.

I guess there are some changes in the monitor code missing, but I'm not an expert here. I would recomment @TheNeikos implements this, though I already have some basic commits which can be used to continue work on this. They are pushed into my repo on "add-method_for_setting_background". Feel free to create a PR based on them!

PR #490, opened 2014-11-24
Add method for setting background

matthiasbeyer:
This PR adds support for setting background via the API.

Targets #438

### DIFF
--- a/src/compositor/module.c
+++ b/src/compositor/module.c
@@ -277,8 +277,10 @@ blit_duck_on_monitor(
     }
     ws_log(&log_ctx, LOG_DEBUG, "Copying into monitor with name: %s",
             monitor->current_mode->mode.name);
-    ws_buffer_blit((struct ws_buffer*) monitor->buffer,
-            (struct ws_buffer*) duck);
+    if (ws_monitor_set_bg_from_buffer(monitor, duck) != 0) {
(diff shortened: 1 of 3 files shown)


neithernut:
Will probably be moved into the egl renderer or something...

matthiasbeyer:
Could this codebase be reused maybe? So can we merge it?

neithernut:
I'm not _quite_ sure whether this should go into the monitor or the egl rendering target... (subspaces with custom backgrounds!!!!11!!eleven)

matthiasbeyer:
I just re-checked the code in this PR. It works with the `ws_monitor` interface, so if the egl stuff implements the interface, this should all work.

We might refactoring this PR, though, because we can set object attributes now way more elegant (the code for the actual command _could_ be emitted)...

neithernut:
Setting an attribute is not sufficient, since you actually have to _do_ something to load the image.
Nevertheless, `ws_buffer_blit` will vanish, so you'll have to rewrite the command (preferably by loading the image into a texture).

matthiasbeyer:
So, I think I will wait until the infrastructure for this is up.

neithernut:
That's excactly what I told you back when I set this "on hold" in the first place.

Issue #495, opened 2014-11-25
Should `ws_object_getref()` lock internally?


matthiasbeyer:
I guess `ws_object_getref()` should not lock internally. We may already have a write/read locked object when doing getref, so maybe it is not required to lock the object in `ws_object_getref()` OR we should provide a function for getting a ref explicitely without doing a lock before incrementing the ref counter...

What do you think?

neithernut:
The `refcnt` member is locked by a _seperate_ lock. If `getref` conflicts with the locking state of the _object_, then something is broken.

Note: the `refcnt` member if currently read-write-locked. I think it's safe to discard this model and use a mutex instead for ref-counting. This makes more sense since `getref` and `unref` bot read and write the refcount.

matthiasbeyer:
Jup, ... sounds good. Feel free to do this! :smile:

neithernut:
Meh.

PR #497, opened 2014-11-26
Fix refcounting in object

neithernut:
Previously, we had a lock for refcounting but didn't use it _at all_.
This commit introduces proper recount-locking and thus fixes a _lot_ of
potential issues.

Targets #495.


matthiasbeyer:
ACK.

matthiasbeyer:
Related to #465

PR #498, opened 2014-11-26
Fix wayland client interaction

TheNeikos:


src/compositor/wayland/buffer.c
```diff
@@ -334,3 +334,11 @@ end_access(
     wl_shm_buffer_end_access(shm_buffer);
 }
 
+void
+ws_wayland_buffer_release(
+    struct ws_wayland_buffer* self
+) {
+    struct wl_resource* res = ws_wayland_obj_get_wl_resource(&self->wl_obj);
```
matthiasbeyer, line 348: Could this possibly be `NULL`?

src/compositor/wayland/surface.c
```diff
@@ -337,6 +338,14 @@ surface_frame_cb(
     uint32_t callback
 ) {
     //!< @todo: implement
+
+    struct ws_surface* surface = ws_surface_from_resource(resource);
+    surface->frame_callback = ws_wayland_client_create_resource(client,
+            &wl_callback_interface, 1, callback);
```
matthiasbeyer: Please add more indentation here.

src/compositor/wayland/buffer.c
TheNeikos, line 348: Yes, but what state should the application go if the buffer is NULL? There is no way to recover from that.

neithernut:
I'll assume rhis is not WIP any more.

src/compositor/keyboard.c
```diff
@@ -239,6 +239,31 @@ ws_keyboard_send_key(
...
+    uint32_t locked;
+    locked = xkb_state_serialize_mods(self->xkb->state, XKB_STATE_LOCKED);
+    uint32_t group;
+    group = xkb_state_serialize_layout(self->xkb->state,
+                                        XKB_STATE_LAYOUT_EFFECTIVE);
+    // ws_log(&log_ctx, LOG_DEBUG, "Direction : %d", direction);
+    // ws_log(&log_ctx, LOG_DEBUG, "Depressed : %x", depressed);
+    // ws_log(&log_ctx, LOG_DEBUG, "Latched : %x", latched);
+    // ws_log(&log_ctx, LOG_DEBUG, "Locked : %x", locked);
+    // ws_log(&log_ctx, LOG_DEBUG, "Group : %x", group);
```
neithernut: Those should either be removed or made real calls (instead of comments).

TheNeikos: You don't want them to be real calls, so I will remove them.

matthiasbeyer: Maybe you can change them to `ws_dbg()` calls? Then, they will only be included in a debug build...

neithernut: A removal would be preferrable.

TheNeikos:
Okay, I fixed all the comments

src/compositor/keyboard.c
```diff
@@ -239,6 +239,26 @@ ws_keyboard_send_key(
         return;
     }
 
+    enum xkb_key_direction direction;
+    direction = state == WL_KEYBOARD_KEY_STATE_PRESSED ? XKB_KEY_DOWN : XKB_KEY_UP;
```
matthiasbeyer: Uuhh... I'm not sure about the binding here. Can you please add parentheses around the condition?

src/compositor/keyboard.c
```diff
@@ -239,6 +239,26 @@ ws_keyboard_send_key(
...
+
+    uint32_t depressed;
+    depressed = xkb_state_serialize_mods(self->xkb->state, XKB_STATE_DEPRESSED);
+    uint32_t latched;
+    latched = xkb_state_serialize_mods(self->xkb->state, XKB_STATE_LATCHED);
+    uint32_t locked;
+    locked = xkb_state_serialize_mods(self->xkb->state, XKB_STATE_LOCKED);
+    uint32_t group;
+    group = xkb_state_serialize_layout(self->xkb->state,
+                                        XKB_STATE_LAYOUT_EFFECTIVE);
```
matthiasbeyer: Can you please move the variable declarations up and the variable settings down and align them? Like so:

``` c
uint32_t ae;
uint32_t b;

ae = foo();
b  = bar();
```

matthiasbeyer: (Would result in more readable code ... at least for me)

src/compositor/wayland/buffer.c
```diff
@@ -120,6 +121,8 @@ end_access(
  *
  */
 
+static struct ws_logger_context log_ctx = { .prefix = "[Compositor/Surface]" };
```
matthiasbeyer: Missing whitespace after `]` in string!

TheNeikos:
This has been changed again

neithernut:
ACK.

PR #502, opened 2014-11-26
Add surface roles

TheNeikos:


neithernut:
ACK.

src/compositor/wayland/surface.c
```diff
@@ -453,3 +458,22 @@ resource_destroy(
...
+    }
+
+    if (self->role && self->role != role) {
+        return -1;
+    }
+
+    self->role = role;
+
+    return 0;
+
```
matthiasbeyer: Please remove this blank line.

src/compositor/wayland/surface.c
```diff
@@ -453,3 +458,22 @@ resource_destroy(
...
+int
+ws_surface_set_role(
+    struct ws_surface* self,
+    struct wl_interface const* role
+) {
+    if (!self || !role) {
+        return -1;
+    }
+
+    if (self->role && self->role != role) {
```
matthiasbeyer: A `ws_surface` is a `ws_object`, isn't it? So it should be read-locked here... I guess!

src/compositor/wayland/surface.c
```diff
@@ -453,3 +458,22 @@ resource_destroy(
...
+        return -1;
+    }
+
+    self->role = role;
```
matthiasbeyer, line 492: ... and write-locked here.

neithernut, line 492: **IF** anything, the whole thing has to be write-blocked. You get deadlocks otherwise.

matthiasbeyer, line 492: So I guess it would be feasible to state this in the documentation of this function.

TheNeikos:
Please do tell if I can do a read lock and then a write lock

matthiasbeyer:
On the same object? I don't think so, but I guess you should wait on @neithernut s answer here.

matthiasbeyer:
Why would that be neccessary? A write-lock includes a read-lock!?

TheNeikos:
Check the code, but I suppose I can do a write lock from the beginning.

TheNeikos:
Okay, I changed the code a bit, please check that I do the locking/unlocking correctly.

src/compositor/wayland/surface.c
```diff
@@ -453,3 +458,27 @@ resource_destroy(
...
+) {
+    int ret = -1;
+    ws_object_lock_write(&self->wl_obj.obj);
+
+    if (!self || !role) {
+        goto away;
+    }
+
+
+    if (self->role && self->role != role) {
```
matthiasbeyer: What about merging this into the upper one?

At least the doubled blank line above should be removed.

src/compositor/wayland/surface.c
```diff
@@ -453,3 +458,27 @@ resource_destroy(
...
+        goto away;
+    }
+
+    self->role = role;
+    ret = 0;
+away:
```
matthiasbeyer: You really wanna call this "away"? :smile:

matthiasbeyer:
The locking looks sane for me.

---

Edit: And btw I don't see why you would need a read lock here anyways...

TheNeikos:
Okay, changed a few things

matthiasbeyer:
ACK.

neithernut:
Needs rebase.

TheNeikos:
Rebased

neithernut:
ACK.

PR #506, opened 2014-11-26
Add stack based initialization for EGL buffers

neithernut:
Makes EGL buffers initializable directly.


matthiasbeyer:
Looks good.

PR #509, opened 2014-11-27
Cleanup user documentation format

matthiasbeyer:
This PR tries to cleanup the user documentation format, so it looks nice.

Feel free to suggest things.

### DIFF
--- a/doc/user/waysome.tex
+++ b/doc/user/waysome.tex
@@ -1,4 +1,4 @@
-\documentclass{book}
+\documentclass[12pt,a4paper]{book}
 
 %
 % Packages


neithernut:
I would have defered the formatting until 2015...
ACK for a4paper/12pt.

neithernut:
The section numbering is odd btw.

matthiasbeyer:
What do you mean with "odd"?

I'd not add more formatting stuff, as you said this should be done in january or so...

neithernut:
Yep, but I can still add a note, so we remember it in January.
By odd, I mean that they always start with `0.`.

matthiasbeyer:
That's indeed odd. If you want, I will investigate asap. But I guess it is not that urgent...

neithernut:
It's not urgent.

PR #513, opened 2014-11-28
Ignore event spams of keys we already have

neithernut:
This PR fixes a bug where keeping a key down for a while causes the DAG not to be traverses correctly.

### DIFF
--- a/src/input/hotkeys.c
+++ b/src/input/hotkeys.c
@@ -118,9 +118,27 @@ ws_hotkeys_eval(
     // get the key code
     uint16_t code = ev->code;
 
-    // insert the event into the track list
+    // insert the event into the track list, if it's not yet in the event list
     {
         struct input_event* buf;
+
+        // check whether the event already is accounted for
+        struct input_event* const data = ws_hotkeys_ctx.events.data;
+        buf = data + ws_hotkeys_ctx.events.size;
+        while (buf-- > data) {
+            if (buf->code != code) {
+                // event not of interest
+                continue;
+            }
+            if (buf->value != ev->value) {
+                // the button was release before
+                break;
+            }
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Will test, stay tuned!

src/input/hotkeys.c
```diff
@@ -118,9 +118,27 @@ ws_hotkeys_eval(
...
+            if (buf->code != code) {
+                // event not of interest
+                continue;
+            }
+            if (buf->value != ev->value) {
+                // the button was release before
+                break;
+            }
+            // we consume the event but we'll _not_ store it or iterate the DAG
+            return empty_eventlist();
```
matthiasbeyer, line 138: Wouldn't it result in less code if you'd do a `if (buf->code == code) { return ... }` here and remove the `continue` therefor?

neithernut, line 138: The extra level of intendation would make the documentation in line 137 a pain in the ass.

matthiasbeyer, line 138: Meh. Could be fixed using a multi-line arg, but well.. I won't argue about that.

matthiasbeyer:
ACK.

matthiasbeyer:
**STOP!**

Not ACK. Press a key twice crashes the system.

neithernut:
May this be related to #514? (Will incluse a fix in this PR btw.)

neithernut:
Please check whether there are still problems.

TheNeikos:
I can't reproduce it

TheNeikos:
#515 is still open, do you think that is related?

neithernut:
Issue #515 _should_ now be resolved (I only tested it briefly).

matthiasbeyer:
ACK.

PR #516, opened 2014-11-29
Fix exec

neithernut:
Our exec code had a few issues.
For example, the effective user id and group id were not reset, which is generally bad. Resettign them means that we can solve the whole need-to-be-in-"input"-group-issue by setting the set-group-ID bit on installation and setting the group of the executable to "input", whithout having forked processed be able to read from `/dev/input/*`.


TheNeikos:
Indeed, ack

matthiasbeyer:
ACK.

PR #533, opened 2014-12-01
Add egl rendering target

neithernut:
This PR introduces a rendering target for EGL buffers.
It's intended to replace the blitting code we have atm through it's blitting method.


neithernut:
The last commit adds a MD file describing how I intend to handle buffers/surfaces. At the end, it roughly explains two options to do this, technically. Feel free to comment, especially on the two options suggested.

src/compositor/buffer/egl_target.h
```diff
@@ -0,0 +1,134 @@
...
+ * @extends ws_egl_buffer
+ */
+struct ws_egl_target {
+    struct ws_egl_buffer egl;   //!< @protected base class
+    struct ws_set surfaces;     //!< @protected shell surfaces blitted on here
+    struct ev_async watcher;    //!< @protected watcher to use for updates
+};
+
+/**
+ * Variable which holds the type information about the ws_egl_targte type
```
matthiasbeyer: Typo.

matthiasbeyer:
I think I'm fine with this, but I'd like you to wait for @TheNeikos comment.

neithernut:
@TheNeikos reviewed it and said it was feasible, we will, however, perform some inqueries on the subject and give me more feedback. I _may_ postpone implementation of the (shared) texture until then.

neithernut:
Mh... some header is missing, ...
Anyways, from this point on, I think I'll have to rely on advise from @TheNeikos  once again. (texture binding etc)
Furthermore, I'll probably split this PR into two.

neithernut:
One question which has yet to be solved is whether binding a texture is thread safe or whether OpenGL needs to be locked globally.

neithernut:
The entire texture area stuff was moved to another PR, which blocks this one.

neithernut:
Turns out this is not needed.

Issue #534, opened 2014-12-01
Pressing hotkeys really slow does not fire a transaction


matthiasbeyer:
Might be related to #532 

Pressing "strg-alt-q" really slowly does not fire the registration which is mapped on these keys, whereas pressing them fast executes it.

Feel free to add appropriate prio.

neithernut:
In the future, **do not file an issue whithout noting anything about potentially missing prequisites** or a working config to reproduce the bug, for that matter!

Issue #532, opened 2014-12-01
Pressing one key after another fires a event


matthiasbeyer:
When registering "strg-alt-q" on an event, "strg", "alt", "event" executes this transaction, too.

Feel free to add appropriate prio.

PR #537, opened 2014-12-01
Fix hotkey evaluation

neithernut:
This commit fixes a bug which causes the hotkey subsystem accept
successive key-presses instead of keeping keys pressed.

Target #532 and #534.

### DIFF
--- a/src/input/hotkeys.c
+++ b/src/input/hotkeys.c
@@ -94,6 +94,7 @@ ws_hotkeys_init(void) {
 
 
     res = ws_hotkey_dag_init(&ws_hotkeys_ctx.root);
+    ws_hotkeys_ctx.state = NULL;
     if (!res) {
         goto cleanup;
     }
@@ -259,6 +260,8 @@ static struct wl_array
 eventlist_reset(void) {
     struct wl_array buf = ws_hotkeys_ctx.events;
     ws_hotkeys_ctx.buttons_pressed = 0;
+    ws_hotkeys_ctx.state = NULL;
+
     wl_array_init(&ws_hotkeys_ctx.events);
     return buf;
 }
--- a/src/input/hotkeys.c
+++ b/src/input/hotkeys.c
@@ -115,30 +115,18 @@ struct wl_array
 ws_hotkeys_eval(
     struct input_event* ev
 ) {
+    // ignore repeated or unknown events
+    if ((ev->value < 0) || (ev->value > 1)) {
+        return empty_eventlist();
+    }
+
     // get the key code
     uint16_t code = ev->code;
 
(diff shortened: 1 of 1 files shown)


neithernut:
I managed to correct the _wrong_ condition...
Please review now.

neithernut:
Ahrg. Fuck it. I'll nitify you when it's ready.

neithernut:
Mh... I found that even the hotkey never gets added with my testing config...
Maybe that's why I had a hard time spotting the bug...
(glaring at @matthiasbeyer)

---

So **now** I'm pissed. With the testing config you gave me, `ws_action_manager_process()` only get's called **once**, which a transaction which has **no flags set** (on the current `develop`). Obviously the DAG contains no nodes and no event will be emitted nor will any of the potentially problematic code in `ws_hotkeys_eval()` be triggered!
Are there **any** unmerged PRs you forgot to tell me about, @matthiasbeyer? Because obviously I can't reproduce a bug if the prequisites are not met!

I wasted a complete hour trying to reproduce it, given my low levels of concentration today!

matthiasbeyer:
I'm sorry, but on the latest develop with the config I gave you, the `ws_action_manager_process()` function gets called twice for me, each time with a transaction which has at least one flag set.

I actually tried this on the latest develop with the configuration file I gave you:
- Start waysome with log level 8 in cgdb
- Put break point on `ws_action_manager_process()`
- Run
- Step to the point where the transaction flags are retreived
- Print these flags (register flag)
- Continue (which breaks at the `ws_action_manager_process()` function again)
- Step to the point where the transaction flags are retreived
- Print these flags (register flag and execute flag)

@TheNeikos and me debugged this yesterday and the dag was built up fine... as far as I can remember...

neithernut:
When I check out current the `develop` branch, do a clean build, set a breakpoint at `ws_action_manager_process()` and run the whole thing, I see excactely **one** transaction.
Do you have **any** changes lying around which could have an influence on the result!?

neithernut:
In the end, it was my fault.
I had a config lying around in my home directory, which was processed by waysome.
Since the deserializer didn't set **any** flags, I assumed the deserializer was broken. Hence the confusion.

I'm terribly sorry for yelling at people. Next time, I'll try to channel my frustration in another way.

neithernut:
This finally works.

matthiasbeyer:
I can execute my testing transaction with strg-alt-q if capslock is enabled, but I guess this is another issue.

I can not reproduce the bug, so this is a happy **ACK**!

matthiasbeyer:
For the record: Capslock is not a problem here, as we watch out for the key codes, not the actual "letter" or something.

PR #543, opened 2014-12-02
Initialize command and flag fields of transactions

neithernut:
This commit adds initialization codes for fields which were left
uninitialized before, potentially leading to segfaults.


matthiasbeyer:
ACK. (so far...)

neithernut:
No more changes...

PR #544, opened 2014-12-02
Fix the execution method of the context

neithernut:
evecve & co. get a NULL-terminated list with the first "argument" being
the command itself (it's a convention). The original code had a few
issues (e.g. it failed to accept commands without arguments). This
pathch fixes those issues by nearly completely rewriting the function.

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -189,37 +189,41 @@ func_exec(
         return -EINVAL;
     }
 
-    struct ws_string* strcmd    = ws_value_string_get(&stack->string);
-    const char* cmd             = ws_string_raw(strcmd);
-
-    stack++; // because `cmd` is extracted now
-
-    size_t n;
-    bool iter = true;
+    size_t n = 0;
     enum ws_value_type iter_type;
 
     // count number of arguments and do type checking meanwhile
-    for (n = 0; iter; n++) {
-        iter_type = ws_value_get_type(&stack[n].value);
-        if (iter_type != WS_VALUE_TYPE_STRING) {
-            return -EINVAL;
-        }
-
-        iter = iter_type != WS_VALUE_TYPE_NONE;
+    while ((iter_type = ws_value_get_type(&stack[n].value)) ==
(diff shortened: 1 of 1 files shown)


neithernut:
Incidently, this allows you to start a `weston-terminal` via a transaction.

src/context.c
```diff
@@ -189,37 +189,41 @@ func_exec(
...
     enum ws_value_type iter_type;
 
     // count number of arguments and do type checking meanwhile
-    for (n = 0; iter; n++) {
-        iter_type = ws_value_get_type(&stack[n].value);
-        if (iter_type != WS_VALUE_TYPE_STRING) {
-            return -EINVAL;
-        }
+    while ((iter_type = ws_value_get_type(&stack[n].value)) ==
+            WS_VALUE_TYPE_STRING) {
```
matthiasbeyer, line 197: Could you possibly use the `ITERARTE_ARGS_TYPE()` macro here?

src/context.c
```diff
@@ -189,37 +189,41 @@ func_exec(
...
 
-        iter = iter_type != WS_VALUE_TYPE_NONE;
+    // check whether there is a command (argument 0)
+    if ((iter_type != WS_VALUE_TYPE_NONE) || !n) {
+        return -EINVAL;
     }
 
-    const char* args[n];
+    // allocate room for a NULL-terminated list of args
+    const char* args[n+1];
```
matthiasbeyer: Please add spaces around the `+` here.

src/context.c
neithernut, line 197: Mh... I have to rewrute a bit, but... yes.

---

Turns out it doesn't make sense. The macro is used to iterate through arguments of some type, but it tries to extract the arguments from the value, which will result in unneccessary function calls (we really only want to count here).
I'd oppose using the macro.

matthiasbeyer, line 197: Allright!

matthiasbeyer:
ACK.

Issue #492, opened 2014-11-24
Add `istype` command


matthiasbeyer:
I think we need a command "isnil" to check whether an object on the stack is `NIL` or not.

I don't know whether this should go into an own file (`src/command/nil.*` for example), but I guess we need it, right?

neithernut:
A way to check for the type of a value would be preferable.

matthiasbeyer:
Hm. We have already "hastypename" for object related once, but I guess we need also one for the value types, yes. I will rewrite this issue for it.

xVin:
May I do this?

neithernut:
Yep.

neithernut:
For clarification: the command should check the _value_'s type (got confused there myself).

xVin:
Is it ok to put this into command/object.c or should it get its own file?

neithernut:
I think a new file wouldl be preferable. Maybe `util` or something...

matthiasbeyer:
I'm sorry I confused you guys...

PR #546, opened 2014-12-02
Add is_type command, is_instance_of command

xVin:
targets #492

### DIFF
--- a/src/command/object.c
+++ b/src/command/object.c
@@ -152,3 +152,31 @@ ws_builtin_cmd_has_attr(
     return 0;
 }
 
+int
+ws_builtin_cmd_is_instance_of(
+    union ws_value_union* args
+) {
+    if ((args[0].value.type != WS_VALUE_TYPE_OBJECT_ID) ||
+        (args[1].value.type != WS_VALUE_TYPE_OBJECT_ID)) {
+        return -EINVAL;
+    }
+
+    struct ws_object* obj = ws_value_object_id_get(&args[0].object_id);
+
+    if (ws_value_object_id_get(&args[1].object_id)) {
+        return -EINVAL;
+    }
+
+    ws_object_type_id* id = ws_value_object_id_get(&args[1].object_id)->id;
+
+    bool b = ws_object_is_instance_of(obj, id);
+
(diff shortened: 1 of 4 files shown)


neithernut:
Mh... I think this command should be renamed into `is_instance_of` or something, as you need two objects for this operation. You can perform a typecheck usng a type-_string_ using `ws_object_has_type`. If you want to, you cann add such a command in this PR.

In the end, this command is not what was requested, but we may need it either way, so this is actually useful.

xVin:
I must have gotten confused somewhere along the way o.o
I will add the is_type command asap.

xVin:
No, wait. I did this because the hastypename command already exists and I wondered what else the is_type command could be supposed to do then.

Edit: I renamed this command anyway, but I guess I need some clarification on what is_type should be doing. Re-reading the issue, it should do typechecking for values only?

neithernut:
I think we just forgot to close that issue...
Sorry...

xVin:
Sure? Because the hastypename command already existed when the issue was opened.

neithernut:
Er.... year, I got confused here...
We _do_ have a command to check an object's typename via API. What's actually missing is a command for checking the type of a `ws_value` itself.

xVin:
Ok so I'll just add that now :)

src/command/object.c
```diff
@@ -151,3 +151,26 @@ ws_builtin_cmd_has_attr(
...
+ws_builtin_cmd_is_instance_of(
+    union ws_value_union* args
+) {
+    if ((args[0].value.type != WS_VALUE_TYPE_OBJECT_ID) ||
+        (args[1].value.type != WS_VALUE_TYPE_OBJECT_ID)) {
+        return -EINVAL;
+    }
+
+    struct ws_object* obj = ws_value_object_id_get(&args[0].object_id);
+    ws_object_type_id* id = ws_value_object_id_get(&args[1].object_id)->id;
```
matthiasbeyer, line 170: I'm not sure, but I think this function could possibly return `NULL`, resulting in a segfault here. Check the return value please!

matthiasbeyer:
Besides the one annotation this looks good!

ACK.

xVin:
Update.

neithernut:
ACK.

PR #548, opened 2014-12-03
Rewrite func_log()

matthiasbeyer:
This PR completely rewrites the func_log() function to be able to
parse `ws_value_*` types into their string representation and log them.

Pair-programmed-with: ~~Marcel Müller com2040@gmail.com~~ @TheNeikos 
Suggested-by: ~~Marcel Müller com2040@gmail.com~~ @TheNeikos

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -27,7 +27,9 @@
 
 #include <errno.h>
 #include <ev.h>
-#include <stdint.h>
+#include <math.h>
+#include <stdlib.h>
+#include <string.h>
 
 #include "command/util.h"
 #include "compositor/cursor.h"
@@ -37,7 +39,9 @@
 #include "input/hotkeys.h"
 #include "objects/object.h"
 #include "objects/string.h"
+#include "util/arithmetical.h"
 #include "util/exec.h"
+#include "util/string.h"
 #include "values/string.h"
 #include "values/union.h"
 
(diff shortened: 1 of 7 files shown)


matthiasbeyer:
Q: Shall I add the appropriate documentation in the user doc?

---

Note: I tested this transaction:

``` json
{
    "TYPE": "transaction",
    "UID": 1,
    "FLAGS": {
        "EXEC": true
    },
    "CMDS": [
        {"push": [true]},
        {"call": [{"pos":0}, "log", "[CONTEXT] ", 3, "FMT", {"pos": -1}]}
    ]
}
```

and it **works!!!**

matthiasbeyer:
I just tested it with

``` json
{
    "TYPE": "transaction",
    "UID": 1,
    "FLAGS": {
        "EXEC": true
    },
    "CMDS": [
        {"push": [true]},
        {"push": [2]},
        {"push": [34]},
        {"push": [null]},
        {"push": ["ENDE"]},
        {"call": [{"pos":0}, "log", "[CONTEXT] ", 3, "FMT ", {"pos": -5}, " ", {"pos": -4}, " ", {"pos": -3}, " ", {"pos": -2}, " ", {"pos": -1}]}
    ]
}
```

**and it still works!!!** ... I think we're awesome! :smile:

neithernut:
I think the conversion of values into strings should probably go into `ws_value_union` or something (would make things more reusable).

matthiasbeyer:
Please review.

I created a to-string functionality for the `ws_value_union` type, which converts the value instances to strings like so:

``` plain
ws_value        -> "<value>"
ws_value_none   -> "<none>"
ws_value_nil    -> "<nil>"
ws_value_bool   -> "<bool:true>" | "<bool:false>"
ws_value_int    -> "<int:1>" (etc)
ws_value_string -> "<string:'blablubb'>"
```

I would like to add to-string functionality for all `ws_value_*` types, also for
the set, which would result in something like this:

``` plain
"<set:[...]>"
```

as I do not want to convert everything in a set to a string ... would be much to heavy!

Suggestions?

neithernut:
I'd postpone this, as it would depend on functionality in the object.

src/util/string.h
```diff
@@ -53,6 +53,16 @@
...
+/*
+ * Convert a hardcoded int to string
+ */
+
+#define STR_HELPER(x) #x
+
+/**
+ * int to string
+ */
+#define STR_OF(x) STR_HELPER(x)
```
neithernut: This is unnecessarily verbose.

src/context.c
```diff
@@ -161,40 +165,55 @@ static int
...
 
-    return 1;
+    while (--n) {
+        if (fary[n]) {
+            free(ary[n]);
+        }
+    }
+
+    return 0;
 }
```
neithernut, line 212: There is no need for `fary`. Storing the index of the element you were processign when bailing out is sufficient for deinitialization in an error case.

src/util/string.h
matthiasbeyer: What would you suggest?

neithernut: `#define STR_OF(x) #x`

matthiasbeyer: This does not work. You actually need a macro-in-macro construct for this!

neithernut: Ok.

TheNeikos: What about `##x` ?

neithernut:
Is there any reason you do a `snprintf("%s%s%s")` to stringify values? Why not `<string:%s>` etc.?

src/util/string.h
matthiasbeyer: I will try...

matthiasbeyer: nope.

matthiasbeyer:
It was because of the length of the prefix and postfix, I didn't like to hard-code them. But I removed the prefix-postfix-foo anyways, as it was ugly to read in the logs. Hope you acknowledge that!

src/util/string.h
TheNeikos: http://codepad.org/DOG1mhlo I am not sure why it's not doing what you want...?

matthiasbeyer: Hm, I had no parentheses... strange things are strange!

TheNeikos: :P Rule #1 in Macros, never _enough_ parantheses.

matthiasbeyer: This was no. 1 rule for lisp... didn't know it also applies to macros! :laughing:

neithernut:
Mh... come to think of it, given a command like:

```
{"log" : ["Current surface id:", {pos: -1}]}
```

Do you really want this to end up like:

```
<string:Current surface id:><int:4>
```

?

matthiasbeyer:
I actually stated in my previous comment that the code which would result in a log-message like this was _removed_ because I considered it as ugly! Now you have a simple concatenation of the log stuff, so your call would result in

```
Current surface id4
```

(I do not add whitespace)

neithernut:
Yep, I just wanted to poitnt this out. Sorry for the confusion.
Anyways: I'm ok with this (ACK for squashing).

matthiasbeyer:
Done.

neithernut:
I think the alst three commits (https://github.com/matthiasbeyer/waysome/commit/36c17065744267d7921a403ab18ace785f201221, https://github.com/matthiasbeyer/waysome/commit/a9fb35dc3ca444be1216970abced2b4c6d57edba and https://github.com/matthiasbeyer/waysome/commit/b6db5ddadfb801fdbff463b08d438e9383286955) should be squashable, too.

matthiasbeyer:
Squashed some more commits, not entirely as you suggested, but almost.

neithernut:
ACK, but you have to rebase now...

Issue #542, opened 2014-12-02
Add connection manager


neithernut:
The connection manager should hold a set of connections and accept connections via a `ws_socket`.

PR #550, opened 2014-12-05
Add connection manager

matthiasbeyer:
This is my [WIP] proposal for the connection manager. I'd like to have a first review on this.

Targets #542.

### DIFF
--- a/src/connection/config_file.c
+++ b/src/connection/config_file.c
@@ -35,7 +35,7 @@
 #include <unistd.h>
 
 #include "connection/config_file.h"
-#include "connection/processor.h"
+#include "connection/manager.h"
 #include "serialize/json/deserializer.h"
 #include "util/arithmetical.h"
 
@@ -76,25 +76,17 @@ open_config(void);
 
 int
 ws_connection_loadconf(void) {
-    int fd = open_config();
-    if (fd < 0) {
-        return fd;
+    int res = ws_connection_manager_init();
+    if (res < 0) {
+        return res;
     }
 
-    // create the deserializer
-    struct ws_deserializer* deserializer;
-    deserializer = ws_serializer_json_deserializer_new();
-    if (!deserializer) {
-        return -1;
+    res = open_config();
+    if (res < 0) {
+        return res;
     }
 
-    struct ws_connection_processor* proc;
-    proc = ws_connection_processor_new(fd, deserializer, NULL);
-    if (!proc) {
-        return -1;
-    }
-    ws_object_unref((struct ws_object*) proc);
-    return 0;
+    return ws_connection_manager_open_connection(res, true);
 }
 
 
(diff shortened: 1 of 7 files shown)


src/connection/config_file.c
```diff
@@ -88,8 +88,8 @@ ws_connection_loadconf(void) {
         return -1;
     }
 
-    struct ws_connection_manager* proc;
-    proc = ws_connection_manager_new(fd, deserializer, NULL);
+    struct ws_connection_processor* proc;
+    proc = ws_connection_processor_new(fd, deserializer, NULL);
```
neithernut: This should be done via the _real_ connection _manager_.
You'll obviously need a method to do this.

src/connection/manager.c
```diff
@@ -25,5 +25,113 @@
...
+#include "connection/processor.h"
+#include "objects/set.h"
+#include "serialize/deserializer.h"
+#include "serialize/json/deserializer.h"
+#include "serialize/json/serializer.h"
+#include "serialize/serializer.h"
+#include "util/socket.h"
+
+//!< @todo no hardcoded socket?
+#define SOCK_NAME "waysome.sock"
```
neithernut, line 42: The socket name will _always_ be hardcoded (or, at least a default value will be).

neithernut:
You are missing deinitialization of the manager.

src/connection/manager.c
matthiasbeyer, line 42: I will remove the `@todo`.

src/connection/config_file.c
matthiasbeyer: What exactly do you mean by "this"? The whole configuration loading?

neithernut: The contruction of a connection for the config.

matthiasbeyer: You would add this to the scope of this PR? I would consider this as stuff for a second PR, as it is mostly refactoring work...

neithernut: Nope, you just have to exrchange a function call or two.

matthiasbeyer:
@neithernut Please review if this is what you wanted me to do (the latest 2 commits).

src/connection/config_file.c
```diff
@@ -76,25 +76,17 @@ open_config(void);
 
 int
 ws_connection_loadconf(void) {
-    int fd = open_config();
+    int fd = ws_connection_manager_init();
```
neithernut: Please, at least rename `fd` to `res`.

neithernut:
Although the config manager is initialized on config load, you should probably initialize it explicitely in `main()` for good meassure.

matthiasbeyer:
You mean the connection manager, but yes.

neithernut:
ACK.

Issue #553, opened 2014-12-05
`ws_socket_deinit()` missing


matthiasbeyer:
The `ws_socket` type has no deinit functionality.

Blocks #550 .

PR #555, opened 2014-12-05
Add ws_socket deinit functionality

matthiasbeyer:
Targets #553

Something missing here? I'm not sure...

### DIFF
--- a/src/util/socket.c
+++ b/src/util/socket.c
@@ -103,6 +103,15 @@ ws_socket_init(
 }
 
 int
+ws_socket_deinit(
+    struct ws_socket* sock
+) {
+    struct ev_loop* loop = ev_default_loop(EVFLAG_AUTO);
+    ev_io_stop(loop, &sock->io);
+    return (close(sock->fd) != 0) ? -errno : 0;
+}
+
+int
 ws_socket_create(
     char const* name
 ) {
--- a/src/util/socket.h
+++ b/src/util/socket.h
@@ -72,6 +72,16 @@ ws_socket_init(
 );
 
 /**
+ * Deinitialize a socket
+ *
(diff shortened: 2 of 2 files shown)


TheNeikos:
Have you checked how to correctly stop libev in that case?

matthiasbeyer:
Nope, that's one point I should do here, right? I don't know why, can you tell me?

matthiasbeyer:
Please tell me whether the code in the latest commit is semantically correct.

TheNeikos:
That is not correct, you will actually unloop the main loop, which will exit waysome.

TheNeikos:
The correct one to use should be `ev_io_stop`

matthiasbeyer:
I would need to pass a watcher to that function... But I do not have one...?

TheNeikos:
The `io` member of the socket is a watcher

matthiasbeyer:
uuuuuh...

neithernut:
ACK for squashing.

matthiasbeyer:
Done. Into one commit. Hope you're fine with this?

neithernut:
The loop is a singleton. It's unnecessary to store a reference in a socket, but I'm fine with this solution.

matthiasbeyer:
Shall I remove the member?

neithernut:
Yep.

matthiasbeyer:
Where can I get the `loop` from to call `ev_io_stop()` with it then?

neithernut:
The same way you get it initially: `ev_default_loop(EVFLAG_AUTO)`

matthiasbeyer:
Updated (and already squashed).

neithernut:
ACK.

Issue #499, opened 2014-11-26
Refactor string callback of json deserializer


matthiasbeyer:
The string callback contains a bunch of doubled code. Refactor this.

(note for myself)

PR #560, opened 2014-12-06
Refactor: String callback helper added

matthiasbeyer:
This patch adds a helper function in the string callback of the deserializer, to reduce code duplication

Targets #499.


neithernut:
Format string should come first.

matthiasbeyer:
You think so? It is just the format for the logging call...?

neithernut:
Yep, but it's a wide-spread convention to have the format string first. I would rather _not_ violate it.

matthiasbeyer:
Updated.

neithernut:
ACK.

Issue #552, opened 2014-12-05
Set up developer's documentation


neithernut:
At some point we agreed that we want a higher level developer's documentation. @matthiasbeyer and me more or less decided on using MD files in following setup:
- One MD file describing the overall architecture on module/package level, including
  - Short description of each module
  - Description of the interaction between modules
- Each module (`compositor`, `input` etc) should get _one_ MD-file, which:
  - describes the module in more detail
  - describes the interaction between classes
  - provides doxygen-links to the classes described
  - may link to (existing) pages describing some function in detail (e.g. `input/hotkeys.md`)
- `main.c` and `context.{h,c}` _may_ get a MD file on their own.

Obviously, the MD files will be embedded into the doxygen documentaiton, serving as introduction/module/group pages/sections.

This issue only covers creating the files with titles and labels. I'd suggest putting module specific documentation into another PR. If the asignee feels that way, he/she may, however, provide content.

matthiasbeyer:
I fully aggree with you, besides one point: I don't think that the `main.c` needs an own documentation file.

But yes for everything else. The assignee should be careful with the location and how the files get embedded, some of the documentation files which do already exist are not put into the right documentation-group. A fix for this could possibly be covered by the assignee as well...

PR #566, opened 2014-12-06
Setup for developer documentation

matthiasbeyer:
This PR starts the setup for the developer documentation.

Targets #552.


neithernut:
I especially like that you dropped `doc/custom` for `doc/developer` makes things easier to understand for people browsing the tree.

matthiasbeyer:
I did not "drop" it, but created another directory. We still need the `doc/custom` directory later on, for the main page for example.

But yes, I thought this would be a bit better structure then putting everything in `custom`.

neithernut:
Wouldn't it make sense to integrate those into `developer`, too?

---

Mh.. we now have the sectio "waysome" and the sectio "Modules" the first one should be integrated into the latter, imho (at least the module main pages and the MD files which are integrated in the source tree).

matthiasbeyer:
I played around a bit and I guess that would be hard to do... I'm not sure how to do it... so feel free to play around with it and provide patches for this PR, but don't get frustrated!

neithernut:
Working on this, I get the serious feeling that either doxygen or our grouping is seriously broken. Check out the commit on my branch. I _just_ added the MD file to the object group, but suddenly a bastard definition of `ws_object` and `ws_string` turns up in the _logger_. WTF? I get that adding MD files to groups _may_ not be supported (for whatever reason), but instead of simply ignorign the `@addgroup`, doxygen fuck's up the document struture!

---

Suddenly I feel the urge to come up with a better piece of software for documenting, but neither do I have the time nor is it a good idea to postpone this until I would have completed that.
Anyways, regarding personal usage, I'll declare doxygen a member of the "broken piece of shit"-group, where it shares a special place with other Qt-related software.

matthiasbeyer:
... didn't I tell you to _not_ get frustrated? It's definitely not doxygens but our own fault here, I just weren't able to catch it by now...

Doxygen is an awesome tool, we're just not able to use it in the right way by now!

---

I will check out your stuff now...

matthiasbeyer:
After some investigation I found out how to do it. I will provide patches as soon as I wrote them...

matthiasbeyer:
The pages are now subpages of the appropriate documentation module. Anyways, they still get included as top-level pages, which I don't think is what we want. I will try to remove them as top-level pages another time...

neithernut:
Being the only tool usable at all doesn't make it "great".
If you can fuck up the document structure by that extend by _simply_ adding _one_ file to a group/section, the system is busted.

matthiasbeyer:
Calm the **FUCK** down. You just used the wrong command, that's it!

neithernut:
I _am_ cal. I'm just done with tools that try to do more than they should, in a way which constrains the usecases they might be used for, and in a bad way on top of that.
Doxygen is a prime example for such a tool.

Btw: I think I need a hint from you in what way the document structure was enhanced. I don't see any enhancements.

matthiasbeyer:
I'm ready for review now.

neithernut:
ACK.

PR #567, opened 2014-12-06
Fix string refcounting in context.c

xVin:
adds missing fixes in context.c

### DIFF
--- a/src/context.c
+++ b/src/context.c
@@ -192,6 +192,8 @@ func_log(
     if (ctx_str && str && lvl) {
         struct ws_logger_context log_ctx = { .prefix = ws_string_raw(ctx_str) };
         ws_log(&log_ctx, lvl, ws_string_raw(str));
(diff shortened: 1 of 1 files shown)


src/context.c
```diff
@@ -221,6 +223,7 @@ func_exec(
 
     while (n) {
         --n;
+        ws_object_unref((struct ws_object*) &stack[i].string);
```
matthiasbeyer: I'm relatively sure we don't need this.

matthiasbeyer: But not entirely. We should ask @neithernut .

xVin: Yeah wasn't sure there as well.

neithernut: The stack itself unrefs/has to unref the values when popping (which, in turn, the action processor does).

xVin:
Updated.

neithernut:
ACK.

PR #569, opened 2014-12-07
[WIP] Add shell compositing

TheNeikos:
This Pull Requests adds a few missing abstraction layers between a surface, its shell, a monitor and the buffers below it.

This is basically all just one feature, make compositing easy.

### DIFF
--- a/src/compositor/buffer/buffer.c
+++ b/src/compositor/buffer/buffer.c
@@ -211,29 +211,7 @@ ws_buffer_blit(
     struct ws_buffer* dest,
     struct ws_buffer const* src
 ) {
-    void* buf_dst = ws_buffer_data(dest);
-    void* buf_src = ws_buffer_data(src);
-    if (!(buf_dst && buf_src)) {
-        return;
-    }
-
-    //!< @todo use byte-size instead of stride
-    ws_log(&log_ctx, LOG_DEBUG, "Blitting image with dim: %dx%d with bpp:%d",
-            ws_buffer_width(src),
-            ws_buffer_height(src),
-            ws_buffer_bpp(src));
-    int min_x = MIN(ws_buffer_width(dest) * ws_buffer_bpp(dest),
-            ws_buffer_width(src) * ws_buffer_bpp(src));
-    int min_y = MIN(ws_buffer_height(dest), ws_buffer_height(src));
-
-    int stride_dst = ws_buffer_stride(dest);
-    int stride_src = ws_buffer_stride(src);
-    for (int y = 0; y < min_y; ++y) {
-        memcpy(((char*) buf_dst) + (y * stride_dst),
-                ((char*) buf_src) + (y * stride_src),
-                min_x
-        );
-    }
+    ws_buffer_blit_at(dest, src, 0, 0);
 }
 
 void
--- a/src/compositor/buffer/buffer.c
+++ b/src/compositor/buffer/buffer.c
@@ -271,3 +271,18 @@ ws_buffer_blit_at(
     }
 
 }
+
+void
+ws_buffer_clear(
+    struct ws_buffer* buff //!< The buffer to clear
+) {
+    void* buf_data = ws_buffer_data(buff);
+
+    if (!buf_data) {
+        return;
+    }
+
+    int size = ws_buffer_stride(buff) * ws_buffer_height(buff);
+    memset(buf_data, 0, size);
+}
+
--- a/src/compositor/buffer/buffer.c
+++ b/src/compositor/buffer/buffer.c
@@ -235,3 +235,61 @@ ws_buffer_blit(
         );
     }
 }
+
+void
+ws_buffer_blit_at(
+    struct ws_buffer* dest,
+    struct ws_buffer const* src,
+    int const x,
+    int const y
+) {
+    void* buf_dst = ws_buffer_data(dest);
+    void* buf_src = ws_buffer_data(src);
+    if (!(buf_dst && buf_src)) {
+        return;
+    }
+
+    //!< @todo use byte-size instead of stride
+    ws_log(&log_ctx, LOG_DEBUG, "Blitting image with dim: %dx%d at (%d, %d)",
+            ws_buffer_width(src),
+            ws_buffer_height(src),
+            x, y);
+
+    // The maximal width can copy. We substract the y position to account for
+    // displacement /into/ the destination buffer.
+    int max_width =  MIN((ws_buffer_width(dest) - y) * ws_buffer_bpp(dest),
+                         ws_buffer_width(src)  * ws_buffer_bpp(src));
+
+    // How many rows can we copy at max?
+    int max_height = MIN((ws_buffer_height(dest) - x), ws_buffer_height(src));
+
+    int stride_dest = ws_buffer_stride(dest);
+    int stride_src = ws_buffer_stride(src);
+
+    //
+    //  We go into the buffer at the x and y position
+    //     <--------Stride------->
+    //     +---------------------+
+    //     |        ^            |
+    //     |        x            |
+    //     |        v            |
+    //     |<   y  >+---+        |
+    //     |        |###|        |
+    //     |        |###|        |
+    //     |        +---+        |
+    //     +---------------------+
+    //  This is done by adding the corret offset. Since a stride is the full
+    //  length of a row we just multiply the x value with it.
+    //  BPP = Bytes per Pixel. This way we then go to the left in the buffer.
+    //  From then on we just
+    buf_dst = ((char*) buf_dst) + (x * stride_dest) + (y * ws_buffer_bpp(dest));
+
+    // vert = vertical position
+    for (int vert = 0; vert < max_height; ++vert) {
+        // We copy a whole row from one buffer into the other
+        memcpy(((char*) buf_dst) + (vert * stride_dest),
+               ((char*) buf_src) + (vert * stride_src),
+               max_width);
+    }
+
+}
--- a/src/compositor/buffer/buffer.h
+++ b/src/compositor/buffer/buffer.h
@@ -246,5 +246,19 @@ ws_buffer_blit(
 __ws_nonnull__(1,2)
 ;
 
+/**
+ * Blit two buffers together at a position (This copies one into the other)
+ *
+ * @memberof ws_buffer
+ *
+ * @warning do not pass NULL to this function! It will crash!
+ */
+void
+ws_buffer_blit_at(
+    struct ws_buffer* dest, //!< The buffer to copy into
+    struct ws_buffer const* src, //!< The buffer top to
+    int const x, //!< Offset of x into `dest`
+    int const y  //!< Offset of y into `dest`
+);
 #endif // __WS_BUFFER_H__
 
--- a/src/compositor/buffer/buffer.h
+++ b/src/compositor/buffer/buffer.h
@@ -260,5 +260,14 @@ ws_buffer_blit_at(
     int const x, //!< Offset of x into `dest`
     int const y  //!< Offset of y into `dest`
 );
+
+/**
+ * Clears a buffer by writing over it with 0s
+ */
+void
+ws_buffer_clear(
+    struct ws_buffer* buff //!< The buffer to clear
+);
+
 #endif // __WS_BUFFER_H__
 
--- a/src/compositor/cursor.c
+++ b/src/compositor/cursor.c
@@ -202,8 +202,11 @@ ws_cursor_set_position(
     int x,
     int y
 ) {
-    int w = ws_buffer_width((struct ws_buffer*) self->cur_mon->buffer);
-    int h = ws_buffer_height((struct ws_buffer*) self->cur_mon->buffer);
+
+    struct ws_buffer* mon_buffer;
+    mon_buffer = (struct ws_buffer*) ws_monitor_get_active_buffer(self->cur_mon);
+    int w = ws_buffer_width(mon_buffer);
+    int h = ws_buffer_height(mon_buffer);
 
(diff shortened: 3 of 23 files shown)


matthiasbeyer:
What a monster of a PR. No way you can split this PR into several smaller ones? (If it's a huge amount of work, I wouldn't care...)

src/compositor/buffer/buffer.c
```diff
@@ -218,20 +228,40 @@ ws_buffer_blit(
...
-                ((char*) buf_src) + (y * stride_src),
-                min_x
-        );
+
+    buf_dst = ((char*) buf_dst) + (x * stride_dest) + (y * ws_buffer_bpp(dest));
+
+    for (int vert = 0; vert < max_height; ++vert) {
+        memcpy(((char*) buf_dst) + (vert * stride_dest),
+               ((char*) buf_src) + (vert * stride_src),
+               max_width);
```
matthiasbeyer, line 270: Can you please add comments on the lines above? From `int max_width...` until here, please!

src/compositor/buffer/buffer.c
```diff
@@ -218,20 +228,40 @@ ws_buffer_blit(
...
-    int min_x = MIN(ws_buffer_width(dest) * ws_buffer_bpp(dest),
-            ws_buffer_width(src) * ws_buffer_bpp(src));
-    int min_y = MIN(ws_buffer_height(dest), ws_buffer_height(src));
+            x, y);
 
-    int stride_dst = ws_buffer_stride(dest);
+    int max_width =  MIN(ws_buffer_width(dest) * ws_buffer_bpp(dest),
+                         ws_buffer_width(src)  * ws_buffer_bpp(src));
+
+    int max_height = MIN(ws_buffer_height(dest), ws_buffer_height(src));
```
matthiasbeyer: `ws_buffer_height()` returns `int32_t`, you should use this here!

src/compositor/buffer/buffer.c
```diff
@@ -218,20 +228,40 @@ ws_buffer_blit(
...
+
+    int stride_dest = ws_buffer_stride(dest);
```
matthiasbeyer, line 244: Should also be `int32_t`!

src/compositor/buffer/buffer.c
```diff
@@ -218,20 +228,40 @@ ws_buffer_blit(
...
             ws_buffer_width(src),
             ws_buffer_height(src),
-            ws_buffer_bpp(src));
-    int min_x = MIN(ws_buffer_width(dest) * ws_buffer_bpp(dest),
-            ws_buffer_width(src) * ws_buffer_bpp(src));
-    int min_y = MIN(ws_buffer_height(dest), ws_buffer_height(src));
+            x, y);
 
-    int stride_dst = ws_buffer_stride(dest);
+    int max_width =  MIN(ws_buffer_width(dest) * ws_buffer_bpp(dest),
```
matthiasbeyer: Should be `int32_t`!

src/compositor/buffer/buffer.c
```diff
@@ -218,20 +228,40 @@ ws_buffer_blit(
...
     int stride_src = ws_buffer_stride(src);
-    for (int y = 0; y < min_y; ++y) {
-        memcpy(((char*) buf_dst) + (y * stride_dst),
-                ((char*) buf_src) + (y * stride_src),
-                min_x
-        );
+
+    buf_dst = ((char*) buf_dst) + (x * stride_dest) + (y * ws_buffer_bpp(dest));
+
+    for (int vert = 0; vert < max_height; ++vert) {
```
matthiasbeyer, line 266: Considering the above, you should take `int32_t` here as well!

src/compositor/buffer/buffer.c
```diff
@@ -218,20 +228,40 @@ ws_buffer_blit(
...
+        memcpy(((char*) buf_dst) + (vert * stride_dest),
+               ((char*) buf_src) + (vert * stride_src),
```
matthiasbeyer, line 269: I guess the whole casting stuff can be removed if you use `char*` from the beginning, can't it?

src/compositor/buffer/buffer.h
```diff
@@ -246,5 +246,30 @@ ws_buffer_blit(
 __ws_nonnull__(1,2)
 ;
 
+/**
+ * Blit two buffers together at a position (This copies one into the other)
+ *
+ * @note Should be called with ref on argument already aquired!
```
matthiasbeyer: On which argument? Please state more clearly!

src/compositor/cursor.c
```diff
@@ -202,8 +202,11 @@ ws_cursor_set_position(
     int x,
     int y
 ) {
-    int w = ws_buffer_width((struct ws_buffer*) self->cur_mon->buffer);
-    int h = ws_buffer_height((struct ws_buffer*) self->cur_mon->buffer);
+
+    struct ws_frame_buffer* mon_buffer;
+    mon_buffer = ws_monitor_get_active_buffer(self->cur_mon);
+    int w = ws_buffer_width((struct ws_buffer*) mon_buffer);
+    int h = ws_buffer_height((struct ws_buffer*) mon_buffer);
```
matthiasbeyer: I'd prefer one cast instead of two. Can you rewrite this part to only contain one cast?

src/compositor/module.c
```diff
@@ -277,7 +278,7 @@ blit_duck_on_monitor(
     }
     ws_log(&log_ctx, LOG_DEBUG, "Copying into monitor with name: %s",
             monitor->current_mode->mode.name);
-    ws_buffer_blit((struct ws_buffer*) monitor->buffer,
+    ws_buffer_blit((struct ws_buffer*) mon_buffer,
```
matthiasbeyer: Couldn't you remove the cast when acessing the inheritance types (`&mon_buffer->buf` like)?

src/compositor/monitor.h
```diff
@@ -61,7 +61,8 @@ struct ws_monitor {
     bool connected; //!< @public is the monitor connected?
     int id; //!< @public the id of the monitor relative to the fb_dev
 
-    struct ws_frame_buffer* buffer; //!< @public The frame buffer
+    struct ws_frame_buffer* buffer[2]; //!< @private The frame buffers
```
matthiasbeyer: Why two? Please state in the comment...

src/compositor/monitor.c
```diff
@@ -170,17 +170,23 @@ ws_monitor_populate_fb(
...
+            self->current_mode->mode.vdisplay
+    );
+    if (!self->buffer[0] || !self->buffer[1]) {
         ws_log(&log_ctx, LOG_CRIT, "Could not create Framebuffer");
     }
     self->saved_crtc = drmModeGetCrtc(ws_comp_ctx.fb->fd, self->crtc);
 
-    int ret = drmModeSetCrtc(ws_comp_ctx.fb->fd, self->crtc, self->buffer->fb,
+    int ret = drmModeSetCrtc(ws_comp_ctx.fb->fd, self->crtc,
+                             self->buffer[self->active_buff]->fb,
```
matthiasbeyer, line 190: This segfaults whenever the buffer is `NULL`, which was checked before, but there is just a log.

matthiasbeyer, line 190: And yes, I know that `_new()` methods shouldn't return `NULL` as `*alloc()` never should.. but you can't be safe...

src/compositor/monitor.c
```diff
@@ -365,3 +371,31 @@ monitor_cmp(
...
+        ws_log(&log_ctx, LOG_ERR, "could not set the crtc for self %d.",
+                self->crtc);
+    }
+}
+
+void
+ws_monitor_flip_buffers(
+    struct ws_monitor* self //!< The monitor where you flip the buffers
+) {
+    self->active_buff ^= 1;
```
matthiasbeyer, line 397: Maybe a short note why this is sufficient would be appropriate...

src/compositor/monitor.h
```diff
@@ -61,7 +61,8 @@ struct ws_monitor {
...
+    int active_buff; //!< @private the current public framebuffer
```
matthiasbeyer, line 66: Maybe an `unsigend int` would be better here, wouldn't it? Assuming you _always_ will have not more than two buffers, you could use a `uint8_t` here, so we do not waste memory.

Maybe even `uint_fast8_t` can be used here... to optimize things a bit...

src/compositor/wayland/abstract_shell_surface.h
```diff
@@ -42,6 +43,10 @@ struct ws_abstract_shell_surface {
     struct ws_wayland_obj wl_obj; //!< @protected: The associated wayland object
     struct ws_surface* surface; //!< @public: The associated surface
     bool visible; //!< Flag for if the shell is visible
+    int height; //!< The height of the window
+    int width; //!< The width of the window
+    int x; //!< The x position of the window
+    int y; //!< The y position of the window
```
matthiasbeyer, line 52: What about unsigned types here? Maybe even the `uint_fastN_t` types, ... don't know which one you need, but I guess `16` is enough, isn't it?

matthiasbeyer, line 52: Of course maybe some function prototypes need to be changed for this...

src/compositor/wayland/compositor.c
```diff
@@ -157,6 +162,41 @@ ws_wayland_compositor_init(void) {
...
 }
 
+void
+ws_wayland_compositor_flush(void) {
+    //!< @todo Add wl_display_locking
+    struct ws_compositing_event* event;
+    wl_list_for_each(event, &wl_comp_ctx.shells, link) {
+        event->callback(event, event->data);
+        wl_list_remove(&event->link);
+        free(event);
```
matthiasbeyer: Would it be faster to do the `wl_list_remove()` and `free()`  calls in a second iteration, to get the best speed out of the event callback calling loop? (I don't know whether this is micro-optimization,... but maybe it is worth it...?)

src/compositor/wayland/compositor.c
```diff
@@ -157,6 +162,41 @@ ws_wayland_compositor_init(void) {
...
+}
+
+void
+ws_compositing_event_new_and_redraw(
+    struct ws_abstract_shell_surface* surface,
+    void* data
+) {
+    struct ws_compositing_event* event = calloc(1, sizeof(*event));
+    if (!event) {
+        return;
```
matthiasbeyer, line 219: No error return here? Why not?

src/compositor/wayland/compositor.h
```diff
@@ -37,5 +44,30 @@
...
+ * This adds an event to the list of transformations to be done until the next
+ * flush.
+ */
+void
+ws_wayland_compositor_add_event(
+    struct ws_compositing_event* event
+);
+
+/**
+ * Creates a new redraw event and schudeles it for the next update
```
matthiasbeyer, line 66: typo

src/compositor/wayland/surface.c
```diff
@@ -45,6 +46,13 @@
  */
 #define WAYLAND_SURFACE_VERSION (1)
 
+/**
+ * Constants used for the commit system for compositing
```
matthiasbeyer, line 51: Sorry, but I'm not smarter after reading this comment... :disappointed:

src/compositor/wayland/surface.h
```diff
@@ -49,10 +49,17 @@ struct ws_surface {
...
+    struct ws_abstract_shell_surface* parent; //!< @protected the parent
     int32_t x; //!< @public x position of this surface
     int32_t y; //!< @public y position of this surface
     int32_t width; //!< @public width of the surface
     int32_t height; //!< @public height of the surface
+    struct {
+        struct ws_wayland_buffer commit_buf; //!< @private buffer for commit
+        int32_t flags; //!< @private bit flags about the current commit status
+        int32_t x; //!< @public possible x position of this surface
+        int32_t y; //!< @public possible y position of this surface
```
matthiasbeyer, line 61: Maybe you also can use the fast types here...

src/objects/object.c
```diff
@@ -502,7 +502,7 @@ ws_object_cmp(
         return (o1 != NULL) ? -1 : 1;
     }
 
-    if ((o1 == NULL) && (o2 == NULL)) {
+    if (o1 == o2) {
```
matthiasbeyer, line 505: :disappointed:

matthiasbeyer, line 505: I guess you should submit this as separated PR.

matthiasbeyer, line 505: ... And the test fix below, of course..

matthiasbeyer:
I hope you don't hate me because of my 24 annotations... :smile:

src/compositor/wayland/abstract_shell_surface.h
TheNeikos, line 52: They have to be signed -> A window can be outside of a monitor

matthiasbeyer, line 52: You still could use the fast types here...

src/compositor/wayland/compositor.c
TheNeikos, line 219: Where should we put the error?

matthiasbeyer, line 219: Return it.

src/objects/object.c
matthiasbeyer, line 505: You can remove the appropriate patch from the history of the PR, as #570 was opened.

src/compositor/wayland/surface.c
```diff
@@ -45,6 +46,13 @@
...
+ */
+
+#define SURFACE_BUFFER_ATTACHED (1 << 1)
+#define SURFACE_DAMAGED (1 << 2)
```
neithernut, line 55: How about making this enums?

PR #574, opened 2014-12-08
Add missing state reset on event emission

neithernut:
When eimitting an event, we also have to reset the state.


TheNeikos:
ack

PR #575, opened 2014-12-08
Fixup of variable declaration

manuelmessner:
The content of the variable `res` will be returned but it was never
declared (goto case).


manuelmessner:
I quickly checked the other goto calls whether there is a similar problem.
I didn't found another occurence. 
Can be merged.

neithernut:
ACK.

PR #576, opened 2014-12-08
Fix: missing negation in is_instance_of()

xVin:


neithernut:
Yep, we _definitely_ missed that one! ACK.

Issue #563, opened 2014-12-06
Add a new context for the connection


neithernut:
The action manager should be expanded to make the `ws_connection_processor` issuing a transaction availible to the transaction itself. This _may_ be done by storing a reference of the connection inside each transaction. If that reference is unset, a nil value should be placed on the stack.
Blocks #564.

PR #578, opened 2014-12-10
Pass connection context to transaction execution

matthiasbeyer:
My approach on/Targets #563.

### DIFF
--- a/src/action/manager.c
+++ b/src/action/manager.c
@@ -71,7 +71,8 @@ struct {
 static struct ws_reply*
 run_transaction(
     struct ws_transaction* transaction, //!< Transaction to run
-    struct ws_value* context //!< context to push on the stack
+    struct ws_value* context, //!< context to push on the stack
+    struct ws_value* connection_context //!< connection ctx to push on stack
 );
 
 /**
@@ -160,7 +161,7 @@ ws_action_manager_process(
 
         if (flags & WS_TRANSACTION_FLAGS_EXEC) {
             // execute the transaction
-            return run_transaction(transaction, NULL);
+            return run_transaction(transaction, NULL, NULL);
         }
 
         return NULL;
@@ -212,7 +213,7 @@ ws_action_manager_process(
         // now, finally, run the transaction
         struct ws_reply* reply;
         reply = run_transaction(transaction,
-                                &ws_event_get_context(event)->value);
+                                &ws_event_get_context(event)->value, NULL);
         if (reply) {
             ws_object_unref((struct ws_object*) reply);
         }
@@ -317,7 +318,8 @@ ws_action_manager_unregister_transaction(
 static struct ws_reply*
 run_transaction(
     struct ws_transaction* transaction,
-    struct ws_value* context
+    struct ws_value* context,
+    struct ws_value* connection_ctx
 ) {
     struct ws_reply* retval = NULL;
     int res;
@@ -353,6 +355,14 @@ run_transaction(
         ++bottom;
         if (context) {
             ws_value_union_init_from_val(bottom, context);
+        } else {
+            ws_value_union_reinit(bottom, WS_VALUE_TYPE_NIL);
+        }
+
+        // initialize the connection context
+        ++bottom;
+        if (connection_ctx) {
+            ws_value_union_init_from_val(bottom, connection_ctx);
         } else {
             ws_value_union_reinit(bottom, WS_VALUE_TYPE_NIL);
         }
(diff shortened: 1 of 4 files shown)


src/action/manager.c
```diff
@@ -159,8 +161,18 @@ ws_action_manager_process(
         }
 
         if (flags & WS_TRANSACTION_FLAGS_EXEC) {
+            struct ws_value_object_id* ctx = NULL;
+            if (opt_ctx) {
+                ctx = calloc(1, sizeof(*ctx));
+                if (!ctx) {
+                    //!< @todo ... aaahm... fuck! What to do?
```
neithernut: Leave it alone, it should be initialized with `nil`.

src/action/manager.c
```diff
@@ -159,8 +161,17 @@ ws_action_manager_process(
...
 
         if (flags & WS_TRANSACTION_FLAGS_EXEC) {
+            struct ws_value_object_id* ctx = NULL;
+            if (opt_ctx) {
+                ctx = calloc(1, sizeof(*ctx));
+                if (ctx) {
+                    ws_value_object_id_init(ctx);
+                    ws_value_object_id_set(ctx, opt_ctx);
+                } // else ... init with NULL...
+            }
```
neithernut: Why are you allocating memory for the object id on the _heap_?

src/action/manager.c
```diff
@@ -209,10 +220,18 @@ ws_action_manager_process(
             goto cleanup_name;
         }
 
+        struct ws_value_object_id* ctx = NULL;
+        if (opt_ctx) {
+            ctx = calloc(1, sizeof(*ctx));
+            ws_value_object_id_init(ctx);
+            ws_value_object_id_set(ctx, opt_ctx);
+        }
+
```
neithernut: Same as above.

src/action/manager.c
matthiasbeyer: That's indeed a point.

src/action/manager.c
```diff
@@ -159,8 +161,13 @@ ws_action_manager_process(
...
 
         if (flags & WS_TRANSACTION_FLAGS_EXEC) {
+            struct ws_value_object_id ctx;
+            ws_value_object_id_init(&ctx);
+            if (opt_ctx) {
+                ws_value_object_id_set(&ctx, opt_ctx);
+            }
             // execute the transaction
-            return run_transaction(transaction, NULL);
+            return run_transaction(transaction, NULL, &ctx.val);
```
neithernut: You should pass either `NULL` or a nil value here if `opt_ctx==NULL`. The first option would probably be easier to implement.

src/action/manager.c
```diff
@@ -209,10 +216,17 @@ ws_action_manager_process(
...
+        if (opt_ctx) {
+            ws_value_object_id_set(&ctx, opt_ctx);
+        }
+
         // now, finally, run the transaction
         struct ws_reply* reply;
         reply = run_transaction(transaction,
-                                &ws_event_get_context(event)->value);
+                                &ws_event_get_context(event)->value,
+                                &ctx.val);
```
neithernut: Same here.

src/action/manager.c
```diff
@@ -159,8 +161,15 @@ ws_action_manager_process(
         }
 
         if (flags & WS_TRANSACTION_FLAGS_EXEC) {
+            struct ws_value* vctx = NULL;
+            struct ws_value_object_id ctx;
+            ws_value_object_id_init(&ctx);
+            if (opt_ctx) {
+                ws_value_object_id_set(&ctx, opt_ctx);
+                vctx = &ctx.val;
+            }
```
neithernut: You _do_ have this seven lines twice in your code. Maybe you should move them to the beginning of the function (e.g. always initialize `vcxt` and use it if needed).

matthiasbeyer: I know. I think it would be more sufficient to keep them but split this function, as it is already over 100 loc. (without the doubled code nearly 100loc) and it is literally a `if() { /* do much stuff */ } else { /* do other stuff */ }`, which could be easily splitted...

neithernut: You can still split the function _and_ have the code eduplicated:
The interface should not be changed (e.g. thee'll always be an `ws_action_manager_process()`), so I'd pull out the code into two functions (e.g. one for transactions, one for events). Those can _still_ get a pointer to the value as an parameter.
The split might be done in another PR.

matthiasbeyer: So I will clean this up and then submit a follow-up PR to split the function.

neithernut:
ACK for squashing.

neithernut:
ACK, but please trigger the CI once more before merging (the last test is failing again).

PR #581, opened 2014-12-10
Added a run command on build to drone

TheNeikos:


TheNeikos:
btw, this fails because waysome segfaults before and thus pkill exits with 1 instead of 0

.drone.yml
```diff
@@ -6,3 +6,6 @@ script:
     - cmake -DHARD_MODE=ON .
     - make
     - ctest -VV
+    - WAYSOME_LOG_LEVEL=8
+    - src/waysome &
+    - sleep 10s && pkill waysome
```
matthiasbeyer, line 11: Why not `pidof waysome && pkill waysome` ?

.drone.yml
```diff
@@ -6,3 +6,6 @@ script:
     - cmake -DHARD_MODE=ON .
     - make
     - ctest -VV
+    - WAYSOME_LOG_LEVEL=8
+    - src/waysome &
```
matthiasbeyer, line 10: Does it actually work to put the env var and the call on two separate lines?

.drone.yml
matthiasbeyer, line 11: Btw: I would not sleep _10_ seconds! That's way too long to wait for!

neithernut:
We triggered the test after `waysome`'s connection manager got fixed, but the test kept failing.
Is it possible that `waysome` dies because there is no DRM device?

.drone.yml
neithernut, line 11: Killing `waysome` before it even had the possibility to segfault is excactely _not_ what we want here.

neithernut:
Mh... now we have the problem of waysome saying it can't initialize the input module...
Maybe `/dev/input` doesn't exist or is empty?

Issue #580, opened 2014-12-10
Waysome is currently broken due to the ws_connection_manager doing bad things


TheNeikos:
As the title says... lots of segfaults, I fixed one, can't be bothered to fix the others...

neithernut:
@matthiasbeyer introduced it...

PR #582, opened 2014-12-11
Fix connection manager

matthiasbeyer:
Targets #580.

### DIFF
--- a/src/connection/manager.c
+++ b/src/connection/manager.c
@@ -165,7 +165,7 @@ connection_manager_deinit(
     void* dummy
 ) {
     ws_object_deinit(&connman.connections.obj);
-    //ws_socket_deinit(&connman.sock);
+    ws_socket_deinit(&connman.sock);
 }
 
 int
--- a/src/connection/manager.c
+++ b/src/connection/manager.c
@@ -97,7 +97,7 @@ ws_connection_manager_init(void)
     }
 
     res = ws_socket_init(&connman.sock, create_connection_cb, SOCK_NAME);
-    if (res != 0) {
+    if (res != 0 && res != -EADDRINUSE) {
         ws_object_deinit(&connman.connections.obj);
         return res;
     }
--- a/src/connection/manager.c
+++ b/src/connection/manager.c
@@ -96,7 +96,8 @@ ws_connection_manager_init(void)
         return res;
     }
 
-    res = ws_socket_init(&connman.sock, create_connection_cb, SOCK_NAME);
+    // `20` is the hardcoded backlog by now
+    res = ws_socket_init(&connman.sock, create_connection_cb, SOCK_NAME, 20);
     if (res != 0 && res != -EADDRINUSE) {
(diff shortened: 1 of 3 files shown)


src/connection/manager.c
```diff
@@ -97,7 +97,7 @@ ws_connection_manager_init(void)
     }
 
     res = ws_socket_init(&connman.sock, create_connection_cb, SOCK_NAME);
-    if (res != 0) {
+    if (res != 0 && res != -EADDRINUSE) {
```
neithernut, line 101: We shouldn't try to bind to an address twice in the first place. If we do, something else is broken.

neithernut:
I still get console spam about waysome not being able to accept a client, with the code `22` (`EINVAL`).
Might be related to my previous annotation.

matthiasbeyer:
~~Tell me how to reproduce this, please.~~

Just found out...

src/connection/manager.c
matthiasbeyer, line 101: It actually doesn't do this. Remove `/run/user/1000/waysome.sock` after you closed waysome, then try again. Everything is fine then.

Seems that we do not clean up the socket.

matthiasbeyer:
The spamming comes from [here](https://github.com/waysome/waysome/blob/develop/src/util/socket.c#L63), shall we remove this log call (is it intended that this runs `while(true)` like)?

neithernut:
There should be not log message if `accept()` causes `errno` to be `EAGAIN` or `EWOULDBLOCK`.
What do you mean _excactely_ with "`while(true)`-like?".

src/connection/manager.c
neithernut, line 101: Then we should. Or set some flags so bind just takes over the old file.

src/util/socket.c
```diff
@@ -60,7 +60,9 @@ socket_build_connection_cb(
     memset(&addr, 0, len);
     int fd = accept(s->fd, (struct sockaddr*) &addr, &len);
     if (fd < 0) {
-        ws_log(&log_ctx, LOG_ERR, "Could not accept client: %d", errno);
+        if (fd != -EAGAIN && fd != -EWOULDBLOCK) {
+            ws_log(&log_ctx, LOG_ERR, "Could not accept client: %d", errno);
+        }
```
neithernut, line 65: You can merge the `if`s.

matthiasbeyer, line 65: nope. re-read the patch!

matthiasbeyer:
Weird. After removing the file, I get my terminal spammed by the messages you mentioned, because `accept()` returns `-EINVAL`. Running waysome a second time (so, the file already exists), this error doesn't happen again.

Any idea why this happenes?

neithernut:
Note for anybody reviewing this: we did not `listen()`.

neithernut:
ACK for squashing (if there's anything to squash)

neithernut:
ACK.

PR #588, opened 2014-12-11
Add implementation for ws_value_object_id to string

matthiasbeyer:
Related to #467

PR #589, opened 2014-12-11
Add implementation for ws_value_set to string

matthiasbeyer:
My approach for `struct ws_value_set` to string via the `ws_value_union` type.

Related to #467.

### DIFF
--- a/src/values/union.c
+++ b/src/values/union.c
@@ -28,6 +28,7 @@
 #include <errno.h>
 #include <string.h>
 
+#include "values/set.h"
 #include "values/union.h"
 #include "values/value_type.h"
 #include "util/string.h"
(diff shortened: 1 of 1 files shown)


TheNeikos:
Can you test this before merging?

matthiasbeyer:
Actually not, the deserializer is not able to deserialize sets by now.

neithernut:
Would close unmerged, it will take a wile before this can actually _be_ implemented.

matthiasbeyer:
Why exactly? This is just a to-string functionality...

neithernut:
Then try to implement it without a relyable way to serialize _and_ deserialize `ws_object_id`s! That's clearly a prequisite for _this_ functionality. Hence, if we cannot have stringification of object ids yet, we also can't implement it for sets.

matthiasbeyer:
Stringification of object ids is implemented in #588.

neithernut:
If you're ok with implementing serialization without any concrete strategies for deserialization, ... Then I _guess_ it's ok to do it this way.

matthiasbeyer:
The point of this is not serialization! The point of this is only logging, so there is no relation to de/serialization!

neithernut:
If you don't intend to implement a more useful format, then I'll ACK this for squashing

matthiasbeyer:
Depends. What do you think is a "more useful" format? I mean, with the `{ ... }` you immediately see that it is a set, with the adiitional `:N` you see how many elements are in there... I don't want to transform all elements to string! That would result in _way_ more overhead here, and as it is only for log()ing, I don't really see why we should need this complexity here...

neithernut:
Then ACK for squashing.

matthiasbeyer:
Done.

neithernut:
ACK.

Issue #584, opened 2014-12-11
Double type?


matthiasbeyer:
Do we actually need a double type?

I would implement one, if necessary (none of a big deal), but I doubt we need one. If not, this issue will track the progress on removing the code fragments for the double type introduction.

Relates #467

neithernut:
I'd omit it for now. If we ever need doubles, or anyone asks for it, we can still implement it.

matthiasbeyer:
I'll do this.

matthiasbeyer:
Btw: This would also remove the object-attribute-reading/writing code for doubles, still ACK?

neithernut:
Yep.

PR #590, opened 2014-12-12
Remove double support

matthiasbeyer:
This PR removes double support from the object-attribute mechanism as well
as from the deserializer.

Targets/Related to #584.


TheNeikos:
ACK

neithernut:
ACK.

Issue #587, opened 2014-12-11
System for identifying execution errors?


matthiasbeyer:
Sorry for the bad title, but I don't know how to phrase it better.

So, my idea is to introduce a "errno-like" definition, which indicates errors in the command execution, including a failure-string table, for easy and consistent logging as well as (way) more simple creation of return-messages (the ones returned via the API).

My idea would be to map some of the `errno.h` values on strings and aditionally introduce more numbers as indicators, also mapped to strings. After introducing this, a command-function is able to return (for example) `-E_WS_EXEC_FAILED` for indicating that execution of a command failed. This would then be mapped onto a string, for example `"Execution of command failed"`, which can be used for logging (etc) in the calling entity.

Would be beneficial for consistency and solve at least one point from #467 ([this one](https://github.com/waysome/waysome/blob/develop/src/action/manager.c#L387), which is rather complex to solve by now).

What do you think?

neithernut:
For now, we could just make use of `strerror()`.

PR #591, opened 2014-12-12
Add errno-string translation

matthiasbeyer:
Targets #587.

Will be extended later on, to be able to translate custom error numbers
to strings, for better debugging of transactions via the API. So this is
something like a "naive" implementation.


TheNeikos:
ACK

neithernut:
While you're at it: you could add a function which initializes a value union with an error string.

neithernut:
ACK.

Issue #592, opened 2014-12-12
Add drone-status-badge to readme?


matthiasbeyer:
Shall I add the drone status badge to the readme?

Would look like this:

Master:
[![Build Status](http://ci.neikos.me/v1/badge/github.com/waysome/waysome/status.svg?branch=master)](http://ci.neikos.me/github.com/waysome/waysome)

develop:
[![Build Status](http://ci.neikos.me/v1/badge/github.com/waysome/waysome/status.svg?branch=develop)](http://ci.neikos.me/github.com/waysome/waysome)

neithernut:
I already annotated the PR: the README.md will also be shipped. Do we want the badges in a shipped README?

PR #593, opened 2014-12-12
Add status badges for branches in README.md

matthiasbeyer:
Targets #592

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,10 @@
 # Waysome
 
+| Branch  | Status |
+| ------- | ------ |
+| master  | [![Build Status](http://ci.neikos.me/v1/badge/github.com/waysome/waysome/status.svg?branch=master)](http://ci.neikos.me/github.com/waysome/waysome)  |
+| develop | [![Build Status](http://ci.neikos.me/v1/badge/github.com/waysome/waysome/status.svg?branch=develop)](http://ci.neikos.me/github.com/waysome/waysome) |
+
 Note: Waysome is work in progress and not ready or useable yet.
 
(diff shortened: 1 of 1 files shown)


TheNeikos:
ACK

neithernut:
Keep in mind that that README will also be shipped.

matthiasbeyer:
What do you mean with "shipped" in this context?

neithernut:
As in "shipped as part of a source package" (e.g. a zip file).
Btw: this approach also doesn't show the state of the actual version you're viewing but the current version.

matthiasbeyer:
Yes, I am (we are) aware of that.

I don't understand what you mean with your "Btw:...".

neithernut:
If, at any point of time in the future, you browse through the history and get the readme displayed, the badges still indicate the _current_ state and not the state the repository was in when the commit you're viewing was committed or authored.
And even if github is smart enough to do "the right thing" here, other plattforms, which might be used to view the repository (in the future), will not handle this correctly.

TheNeikos:
Tbh that is a non issue, whatever the 'badges' show is not important. It's meant to show how healthy a project is and if it the lastest stage actually builds. It's basically a contract from us to others to make sure that there is always a buildable state.

Now, previous versions do not interest us that much right now. We're not a company with long lasting stable releases and shouldn't need them. Hence why it's not really important that the badges show future versions because they are the only things that matter!

neithernut:
Of course, only future releases matter. But from the point of view of a future self, the current release is a previous one. My point is that you break the concept of "I'm viewing an old release, so _all_ the content I see is from back then." by introducing badges this way.

PR #595, opened 2014-12-12
Moved the heapalloced flag from init to new for ws_named

TheNeikos:


matthiasbeyer:
ACK.

src/objects/named.c
```diff
@@ -101,11 +100,14 @@ ws_named_new(
         return NULL;
     }
 
+
```
neithernut: Why did you insert this blank line?

matthiasbeyer: Seems to be not on purpose...

neithernut:
ACK.

PR #596, opened 2014-12-12
Adds a memset for the ws_value_named on the stack in action/manager

TheNeikos:


matthiasbeyer:
ACK.

Wall of shame?

TheNeikos:
Haha

neithernut:
Did this trigger a bug?
When I wrote this, I think I assumed that the initialization function sets all the values to a sane value. Maybe we should discuss this semantic at some point, but not in this PR.
Anyways, ACK for now.

TheNeikos:
They usually do, but in this case it depended on an input variable.

Issue #597, opened 2014-12-12
Implement `ws_queue` type


matthiasbeyer:
Related to #565.

(As we have nothing to do, no issues are open we can actually do, I open this issue now and assume that it will be required for #565).

matthiasbeyer:
Will be done with @TheNeikos in pair-programming.

PR #598, opened 2014-12-12
Add ws_queue type

matthiasbeyer:
Targets #597.

~~Pair-programmed-with: Marcel Müller com2040@gmail.com~~
Pair-programmed-with: @TheNeikos

### DIFF
--- a/src/objects/queue.c
+++ b/src/objects/queue.c
@@ -145,8 +145,7 @@ bool
 ws_queue_empty(
     struct ws_queue* q
 ) {
-    //!< @todo implement
-    return false;
+    return (q ? wl_list_empty(&q->link) : true);
 }
 
 /*
--- a/src/objects/queue.c
+++ b/src/objects/queue.c
@@ -25,6 +25,9 @@
  * along with waysome. If not, see <http://www.gnu.org/licenses/>.
  */
 
+#include <errno.h>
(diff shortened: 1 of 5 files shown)


neithernut:
_Has_ it to be an object (as in `ws_object`)?

matthiasbeyer:
If you mean the `ws_queue` type: It gives us easy locking and the like.

If you mean the values inside: I assumed that `ws_object`s are of relevance here, as it is written to hold events, if I'm right...

neithernut:
I mean the queue itself.
I do not think that it should be a `ws_object` _just_ because it automatically introduces locking facilities.

TheNeikos:
There is no reason for me for it to _not_ be a `ws_object`. It makes it easier to use and it fits our setup.

src/objects/queue.c
```diff
@@ -0,0 +1,168 @@
...
+
+struct ws_object*
+ws_queue_pop(
+    struct ws_queue* q
+) {
+    if (!q || wl_list_empty(&q->link)) {
+        return NULL;
+    }
+
+    ws_object_lock_write(&q->obj);
```
neithernut: You forgot to check the return value of this call.

neithernut:
ACK for squashing.

neithernut:
ACK.

PR #599, opened 2014-12-13
Fix: Add missing memset() calls

matthiasbeyer:
This PR adds some missing memset() calls when allocating memory on the
stack, which is exaclty the bug which causes a segfault on the latest
develop.


neithernut:
ACK.

PR #608, opened 2014-12-17
Call ws_errno_tostr() in action manager error reply building

matthiasbeyer:
Partly targets #467 and removes the `@todo` from `src/action/manager.c`.
Anyways, the introduced changes may be discussed.


neithernut:
**Much** better. ACK.

PR #615, opened 2014-12-17
Rip logging contexts of compositor apart

matthiasbeyer:
This PR tears the logging contexts of the compositor in smaller pieces
(each part of the compositor has another logging context now).


neithernut:
ACK.

Issue #614, opened 2014-12-17
Add automatic format detection in image loader


neithernut:
We may need to support loading both RGBA and ARGB. Hence, the image loading code should be capable of supporting both formats. For this to be achieved, the following steps have to be taken:
- [x] Add a new field to the mappings in `src/util/egl.{h,c}` for the libpng formats
- [x] Add a new parameter to `ws_image_buffer_from_png()` for passing the format.

Note that the issues which this issue targets _may_ be solved in adifferent way, which I'm currently unable to forsee. Ths, if someone has a better idea, post it!

PR #617, opened 2014-12-17
Add png format mappings

matthiasbeyer:
Targets #614.


neithernut:
Please adjust `ws_egl_fmt_advertise` to only advertise formats for which an egl format-type pair exists.

neithernut:
ACK.

Issue #620, opened 2014-12-18
Add z-index to surfaces


neithernut:
We'll need a z-index at some point.
This also includes a way to set them via the API. Kind of relates to #467.

neithernut:
The API stuff is still missing, but as this has to be done for _all_ the attributes, I'll close this issue now.

PR #623, opened 2014-12-18
Add zindex for ws_surface

matthiasbeyer:
This PR targets #620.

I'm not quite sure whether this is the appropriate approach. Also,
please note that I set the default z index to 1000, which clearly
must be discussed.

### DIFF
--- a/src/compositor/wayland/abstract_shell_surface.h
+++ b/src/compositor/wayland/abstract_shell_surface.h
@@ -54,6 +54,7 @@ struct ws_abstract_shell_surface {
     struct ws_wayland_obj wl_obj; //!< @protected: The associated wayland object
     struct ws_surface* surface; //!< @public: The associated surface
     bool visible; //!< Flag for if the shell is visible
+    int32_t z; //!< @public z index of this surface
 };
 
 /**


neithernut:
If the z-index were an int, we could have a _much_ intuitive default of `0`.

matthiasbeyer:
So shall I dynamically return either a bool (false for a "something did not work") or an int on `get` here?

matthiasbeyer:
... because I don't think we should entirely fail here if this does not work...

neithernut:
This should solve automatically when porting it to the generalized attribute modification and query infrastructure.

matthiasbeyer:
I will rewrite this now with #626 in mind.

neithernut:
ACK for squashing.

matthiasbeyer:
Done.

neithernut:
Would you mind _not_ removing that blank line from `abstaract_shell_surface.c` in your one commit? You're not doing anything in that file anymore, anyways.

TheNeikos:
He will, the Z-index is going to be on the abstract shell, the z index of surfaces is automatical through their parent/child relationship.

matthiasbeyer:
@TheNeikos So I should move the z index to the abstract shell type?

neithernut:
Mh... yep, I didn't consider doing it this way, but if the position is made an offset relative to it's parent (the shell surface), so would the z-index. Since an offset doesn't make sense for the z-index, it would be an attribute of the shell surface only.
TL;DR: Yep, make it an attribute of `ws_abstract_shell_surface`.

neithernut:
ACK for squashing.

matthiasbeyer:
Done.

neithernut:
ACK.

PR #625, opened 2014-12-18
Add object attribute writing

matthiasbeyer:
This is a preparation PR for a HUGE refactoring I'm working on.

Using this, we can refactor the whole abstract shell surface attribute
handling (for example, I'd bet half of the chocolate I will receive on
christmas that there are much more locations we can use this) and
remove quite a big bunch of code in favour of abstraction.

Please review carefully and provide me your suggestions.

### DIFF
--- a/src/objects/object.c
+++ b/src/objects/object.c
@@ -458,6 +458,106 @@ ws_object_attr_read(
     return 0;
 }
 
+int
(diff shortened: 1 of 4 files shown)


neithernut:
ACK.

Issue #601, opened 2014-12-14
Problem with multiple transactions


xVin:
After just testing waysome on my machine @matthiasbeyer and me found out that if two different transaction are added to waysome.json (we used the two transactions from the wiki), we have the following scenario:
- (1) Shutdown, (2) Terminal: Shutdown doesnt work, terminal can be opened (but no input possible, cursor can't be moved anymore)
- (1) Terminal, (2) Shutdown: Segfault on trying to execute any command
- (1) Shutdown: Works
- (1) Terminal: Nothing works, not even CTRL+C

matthiasbeyer:
Can't reproduce:

> (1) Terminal: Nothing works, not even CTRL+C

---

Can reproduce:

> (1) Shutdown: Works

---

Can reproduce:

> (1) Terminal, (2) Shutdown: Segfault on trying to execute any command

(when trying to open the terminal -> segfault) Backtrace:

```
#0  0x000000000040d731 in ws_hotkey_dag_next (node=0x626a80 <ws_hotkeys_ctx>, code=28) at /home/m/dev/c/waysome/src/input/hotkey_dag.c:189
#1  0x000000000040cd33 in ws_hotkeys_eval (ev=0x7fffffffde30) at /home/m/dev/c/waysome/src/input/hotkeys.c:149
#2  0x000000000040e6cf in watch_pointers (loop=0x7ffff7ddaa80, watcher=0x6542d8, revents=1) at /home/m/dev/c/waysome/src/input/input_device.c:246
#3  0x00007ffff7bd05d3 in ev_invoke_pending () from /usr/lib/libev.so.4
#4  0x00007ffff7bd3c0e in ev_run () from /usr/lib/libev.so.4
#5  0x000000000040a100 in ev_loop (loop=0x7ffff7ddaa80, flags=0) at /usr/include/ev.h:828
#6  0x000000000040a45b in main (argc=1, argv=0x7fffffffe108) at /home/m/dev/c/waysome/src/main.c:165
```

matthiasbeyer:
After some investigation:

The segfaults comes from [here](https://github.com/waysome/waysome/blob/develop/src/input/hotkey_dag.c#L189). The cause looks like this:

```
(gdb) p cur.nodes
$42 = {dag = 0x600000005, tab = 0x600000005}
(gdb) p cur.nodes.tab
$43 = (void **) 0x600000005
(gdb) p cur.nodes.tab[0]
Cannot access memory at address 0x600000005
```

[...]

PR #630, opened 2014-12-20
Fix add_roots_for()

neithernut:
During investigation of issue #601, @matthiasbeyer found that, apparently, the function `add_roots_for()` misbehaves. Thi PR is aimed at fixing that function.

Note that this PR contains a rewrite of that function. It appears that that rewrite either breaks some functionality or uncovers another bug. I'd suggest investigating whether the other two fixes resolve the issue, e.g. whether the rewrite can be omitted. That would be a job for @matthiasbeyer as he found the original bug.

### DIFF
--- a/src/input/hotkey_dag.c
+++ b/src/input/hotkey_dag.c
@@ -31,6 +31,7 @@
 
 #include "input/hotkey_dag.h"
 #include "input/hotkey_event.h"
+#include "util/arithmetical.h"
 
 
 #define DAG_TAB_CHILD_NUM_EXP (4)
@@ -370,30 +371,38 @@ add_roots_for(
         return 0;
     }
 
-    // step on which the node is based.
-    int step = DAG_TAB_CHILD_NUM_EXP * tab->depth;
+    // determine the number of nodes by which to extend the tree upwards
+    int_fast8_t step = 0;
+    {
+        // calculate the position of the current node relative to the new root
+        uint_fast16_t diff = ABS((code - tab->start)) >>
+                             (DAG_TAB_CHILD_NUM_EXP * tab->depth);
 
-    // position within the node
-    size_t pos = (code - tab->start) >> step;
+        // we "discard" one level-specific address-part per step
+        while (diff != 0) {
+            ++step;
+            diff >>= DAG_TAB_CHILD_NUM_EXP;
+        }
+    }
+
+    // calculate the "position" of the current node within the new tree
+    size_t pos = tab->start >> (DAG_TAB_CHILD_NUM_EXP * (tab->depth + 1));
 
     // extend the table "upwards", if necessary
-    while ((code < tab->start) || (pos > DAG_TAB_CHILD_NUM)) {
+    while (step-- > 0) {
         // we have to create a new node
         void** tab_node = create_tab_node();
         if (!tab_node) {
             return -ENOMEM;
         }
 
-        size_t old_root_pos = (tab->start >> step);
-
         // put in the new root
-        tab_node[old_root_pos  & (DAG_TAB_CHILD_NUM - 1)] = tab->nodes.tab;
+        tab_node[pos  & (DAG_TAB_CHILD_NUM - 1)] = tab->nodes.tab;
         tab->nodes.tab = tab_node;
-        ++tab->depth;
 
-        // regen step, pos and start
-        step += DAG_TAB_CHILD_NUM_EXP;
-        tab->start = old_root_pos << step;
+        // regen depth, start and pos
+        ++(tab->depth);
+        tab->start &= ~((1 << (DAG_TAB_CHILD_NUM_EXP * (tab->depth + 1))) - 1);
         pos >>= DAG_TAB_CHILD_NUM_EXP;
     }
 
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
Shall I wait until this actually builds?

matthiasbeyer:
Now it segfaults in `src/compositor/buffer/buffer.c`, [here](https://github.com/waysome/waysome/blob/develop/src/compositor/buffer/buffer.c#L175) as the ptr is not valid. 

I'm not sure, though, whether this is already implemented. Assuming from the ptr value, it is not (I'm too lazy to look it up right now).

So it _seems_ that your patches fix it. I will investigate further, though.

matthiasbeyer:
I tried the following configurations:
- shutdown
- shutdown, terminal
- terminal
- terminal, shutdown

They all worked both for shutdown and starting the terminal (while segfaulting because of the false ptr value mentioned above). So I think your patches are sufficient:

**ACK**

This is for the HEAD of the PR! I would like to re-review after you squashed, to be sure everything still works.

neithernut:
**Read the comment I wrote to this PR!**
I **explicitely stated** that the current state introduces new problems but you should test it with **the first two commits only!**

neithernut:
I found a minor bug in `ws_hotkey_dag_remove()`. Apparently, the event is removed from the tree, but not all the nodes that could be removed are, in fact, removed.
I also found the reason why this minor bug didn't show up previously: the insertion method is imperfect. It adds one unnecessary level in the dag node table.
I'm working on fixing those issues.

neithernut:
_Slowly_ getting closer here.

neithernut:
_Now_ this is ready for review and merge (_finally_).

matthiasbeyer:
Bad news:

```
#0  0x000000000040e71d in destruct_tab_node (tab_node=0x4, depth=12 '\f') at /home/m/dev/c/waysome/src/input/hotkey_dag.c:404
#1  0x000000000040e72a in destruct_tab_node (tab_node=0x67da30, depth=13 '\r') at /home/m/dev/c/waysome/src/input/hotkey_dag.c:404
#2  0x000000000040e72a in destruct_tab_node (tab_node=0x6735b0, depth=14 '\016') at /home/m/dev/c/waysome/src/input/hotkey_dag.c:404
#3  0x000000000040e72a in destruct_tab_node (tab_node=0x672100, depth=15 '\017') at /home/m/dev/c/waysome/src/input/hotkey_dag.c:404
#4  0x000000000040e72a in destruct_tab_node (tab_node=0x67cee0, depth=16 '\020') at /home/m/dev/c/waysome/src/input/hotkey_dag.c:404
#5  0x000000000040e090 in ws_hotkey_dag_deinit (entry_node=0x629800 <ws_hotkeys_ctx>) at /home/m/dev/c/waysome/src/input/hotkey_dag.c:144
#6  0x000000000040da6e in hotkeys_deinit (dummy=0x0) at /home/m/dev/c/waysome/src/input/hotkeys.c:264
#7  0x000000000041d8ea in ws_cleaner_run () at /home/m/dev/c/waysome/src/util/cleaner.c:99
#8  0x000000000040adf3 in main (argc=1, argv=0x7fffffffe118) at /home/m/dev/c/waysome/src/main.c:177
```

Steps to reproduce:

```
$ gdb ./src/waysome
run
<hit STRG-ALT-Q>
```

neithernut:
Damn. Well, at least this is easily testable, but strangely, deinitialization _should_ be covered by the tests already, _or is it_?

Note: `depth` _must_ be `<= 4`.

neithernut:
Turns out deinitialization was indeed, completely left out when writing the test cases.

neithernut:
The problem described by @matthiasbeyer was fixed and he also kindly provided updates to the tests. Can I get an ACK?

---

WTF, _now_ the tests start failing.

neithernut:
Some minor fix seemed to have fixed the problem, please test thorougly.

matthiasbeyer:
I did a short test and everything seems to work as expected now. But, I think I'm too tired right now to do a in-depth test and therefor I'd like to postpone this until tomorow?

neithernut:
I'm OK with this, but I'd like to get this merged tomorrow as the bug(s) fixed by this PR are quite nasty.

src/input/hotkey_dag.c
```diff
@@ -403,34 +388,24 @@ create_tab_node(void) {
...
     // iterate over all the nodes
     while (cur_node-- > (void**) tab_node) {
         // destruct children
-        if (depth) {
-            destruct_tab_node(*cur_node, depth);
-        } else {
-            destruct_dag_node((struct ws_hotkey_dag_node*) cur_node);
+        if (depth > 0) {
+            destruct_tab_node(*cur_node, depth - 1);
+        } if (*cur_node) {
```
matthiasbeyer: Is this supposed to be an `if` ? If yes, style fuckup, if not bug!

matthiasbeyer:
ACK from me. Let @TheNeikos check it, though!

neithernut:
@TheNeikos ACK'ed verbally.

Issue #628, opened 2014-12-18
Make attribute getter and setter consider basetype attributes


neithernut:
The getter and setter facilities currently only look for the attribute in the table attached to the type itself but not in the base type(s).

PR #631, opened 2014-12-20
Add attribute reading/writing with basetype checking

matthiasbeyer:
Targets #628.


neithernut:
You don't break the outer loop the moment you found the attribute entry.

neithernut:
ACK for squashing.

neithernut:
ACK.

Issue #633, opened 2014-12-21
`ws_buffer_transfer2texture()' segfaults


matthiasbeyer:
This is because the iteration does not stop at a certain level. (The condition in the basetype iteration is completely wrong here)

Where to stop? At `ws_wayland_obj`?

matthiasbeyer:
After some investigation:
`transfer2texture` gets a `ws_wayland_buffer`, which is _not_ a `ws_buffer`, whereas the comparison with `ws_buffer_type_id` completely fails here...

I'm not sure what _I_ am doing wrong here, or am I doing wrong here at all... We had an issue whether `ws_wayland_buffer` shall be an `ws_buffer` ( #170 ), but it got closed with a comment from @neithernut that it shouldn't. I do not understand why, but trusting his opinion, I do not see how to solve this issue (edit: the basetype-iteration) here in a cleanish way.

Suggestions?

(tested on top of #630, btw)

neithernut:
You have to _not_ use the `ws_wayland_buffer` directly but the `ws_buffer` you get through `ws_wayland_buffer_get_buffer()`!
This is not a bug but malicious usage of the object. I think I'll add a big `@warning` to the datatype, so I can shout out "RTFM" if this happens again.

---

Turns out the idiot who did a direct cast here was me.

PR #634, opened 2014-12-21
Add warning to wayland buffer

neithernut:
...to prevent devs from using it as a `ws_buffer` and filing bug-reports
because the method segfaults.

Should prevent issues like #633 from popping up.


matthiasbeyer:
ACK.

TheNeikos:
ACK

Issue #635, opened 2014-12-22
Sanitize error handling in JSON serializer


neithernut:
The JSON-serializer's error handling consists mostly of TODOs.
Relates to #467.

matthiasbeyer:
@neithernut : Please comment on [this](https://github.com/waysome/waysome/blob/develop/src/serialize/json/serializer.c#L268) what to do? Do I need to free this buffer or does this someone else?

neithernut:
You have to unref it once. (it's ref'd by a `ws_serializer` method)

PR #639, opened 2014-12-22
Add json-serializer error handling

matthiasbeyer:
This PR targets #635.

It only contains logging calls by now, as I'm not sure what to do when
the serialization fails. It should not fail, of course. But if that happens
we could almost do an shutdown, as the serialization should only fail if
no memory is left, and ... yeah... you get it, I guess.

What to do in case of error? Clean up, of course, but what else? Is
logging enough?

### DIFF
--- a/src/serialize/json/serializer.c
+++ b/src/serialize/json/serializer.c
@@ -194,7 +194,7 @@ serialize(
 
         yajl_gen_status stat = yajl_gen_map_open(ctx->yajlgen);
         if (stat != yajl_gen_status_ok) {
-            //!< @todo error opening map, what to do now?
+            ws_log(&log_ctx, LOG_DEBUG, "Error opening main map");
             return -1;
         }
 
--- a/src/serialize/json/serializer.c
+++ b/src/serialize/json/serializer.c
@@ -336,8 +336,7 @@ serialize_reply_value_reply(
 
     // We haven't serialized anything
     // generate the key for the value reply
-    if (gen_key(ctx, VALUE)) {
-        //!< @todo error?
+    if (unlikely(gen_key(ctx, VALUE))) {
         return -1;
     }
 
@@ -347,8 +346,7 @@ serialize_reply_value_reply(
         return -1;
     }
 
-    if (gen_key(ctx, TRANSACTION_ID)) {
-        //!< @todo error?
+    if (unlikely(gen_key(ctx, TRANSACTION_ID))) {
         return -1;
     }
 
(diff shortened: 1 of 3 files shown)


neithernut:
We have to return an error of some sort, so the connection can be shut down by the connection processor.

---

The documentation of `serialize_f` btw sais: "If an internal error occured, a negative error may be returned.". The return value isn't used _yet_, but tht's another TODO (which also exists as a comment/tag in the code)

src/serialize/json/serializer.c
```diff
@@ -492,26 +490,39 @@ serialize_value(
...
+                {
+                    int r = ws_value_set_select((struct ws_value_set*) val,
+                                                 NULL, NULL,
+                                                (int (*)(void*, void const*))
+                                                serialize_object_to_id_string,
+                                                ctx);
+
+                    if (unlikely(r < 0)) {
+                        ws_log(&log_ctx, LOG_DEBUG, "Error serializing set");
+                        //!< @todo abort serialization here?
```
neithernut: Yep, you should abort serialization here.

matthiasbeyer:
The serializer already returns negative values on failure. Maybe they do not tell you anything (`-1`), but sometimes they are like `-EINVAL` or something. We should maybe change the (abstract) interface here, so that some sort of error can be returned (non-errno.h numbers).

neithernut:
I saw that you returned `-1`, I just wanted to emphasize that all that's to be done is
- logging
- returning an error code

Returning an extensive error description is not necessary, since the functions already do logging. _maybe_ we want to return `-EIO` instead of `-1` and _maybe_ we even want to return `-EAGAIN` on some recoverable errors, but that's all I would change in the code base at this point.

matthiasbeyer:
`maybe`? :smile: 

I'm not quite sure that there are _any_ recoverable errors here... I get a `yajl_state_error` if the parser fails and I'm then able to get an error-string describing the error. This does not help very much in manner of "recoverable" errors...

neithernut:
In this case, return `-EIO` instead of `-1`.

matthiasbeyer:
I checked: It makes no sense to return `-1` in any place, all occurences of `-1` in `serialize.c` were replaced by `-EIO` by the latest commit.

Feel free to suggest more changes, etc...

neithernut:
ACK for squashing or merge, if you feel that the history is clean enough.

matthiasbeyer:
Squashed one commit, feel free to merge.

Issue #638, opened 2014-12-22
Add position to abstract shell surface


neithernut:
Title sais it all. Blocks #637.

matthiasbeyer:
I can do this if you want.

PR #643, opened 2014-12-23
Add x/y position to abstract shell surface

matthiasbeyer:
Targets #638.

Anything else to do?


neithernut:
Would make them public and omit the getter and setter. And you have to rebase, of course.

neithernut:
ACK btw.

Issue #637, opened 2014-12-22
Port attributes of abstract shell surface to attribute table


neithernut:
This includes removing teh old getter/setter functions.

matthiasbeyer:
I can do this, if you want.

matthiasbeyer:
Will wait for #638 / #643

matthiasbeyer:
Shall width and height members be part of the `struct ws_abstract_shell_surface`? Should this be done in another PR, maybe?

neithernut:
I'm not yet quite sure about this, would decide this later.

matthiasbeyer:
So I'll submit my attribute-table PR now, but note that it might should wait on a decision on this...

PR #646, opened 2014-12-23
Add attribute table for ws_abstract_shell_surface

matthiasbeyer:
Targets #637.

We _might_ wait with this PR to be merged until we have a decision on
whether the width and height attributes should be part of the
`struct ws_abstract_shell_surface`.


neithernut:
Nah, we can add them later here, too, if necessary.
ACK.

matthiasbeyer:
I guess @TheNeikos stuff will introduce this anyways... so ... feel free to merge!

PR #647, opened 2014-12-26
Remove casts where possible

matthiasbeyer:
This PR removes casts where possible. Please review _carefully_, I did no
double-checking whether I accidentially changed functionality!

Compiled and tested (`make test`) sucessfully!


neithernut:
ACK.

Issue #650, opened 2014-12-27
Sanitize connection processor semantics


neithernut:
As connections last as long as the connection does, and to mitigate some potential race-conditions, connection and disconnection of connection processors should be moved to extra methods.

PR #651, opened 2014-12-27
Add connection connection disconnection methods

neithernut:
Targets #650.
Still missing:
- [x] Locking

### DIFF
--- a/src/connection/processor.c
+++ b/src/connection/processor.c
@@ -218,6 +218,25 @@ ws_connection_processor_start(
     return 0;
 }
 
+int
+ws_connection_processor_close(
+    struct ws_connection_processor* conn
+) {
+    ws_object_lock_write(&conn->obj);
+
+    if (conn->is_init) {
+        conn->is_init = false;
+
+        ev_io_stop(EV_DEFAULT_ &conn->dispatcher);
+        if (conn->serializer) {
+            ev_prepare_start(EV_DEFAULT_ &conn->flusher);
+        }
+    }
+
+    ws_object_unlock(&conn->obj);
+    return 0;
+}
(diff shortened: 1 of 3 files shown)


src/connection/processor.c
```diff
@@ -205,6 +200,39 @@ ws_connection_processor_new(
...
+
+int
+ws_connection_processor_close(
+    struct ws_connection_processor* conn
+) {
+    if (!conn->is_init) {
+        return 0;
+    }
+
+    conn->is_init = false;
```
matthiasbeyer: Sure we want to use this flag for it? Sure we do not want another `is_running` flag?

neithernut: I _could_ rename the flag. But the is_init flag itself doesn't make any sense, so I decided to use it for indicating this state instead.

src/connection/processor.h
```diff
@@ -54,7 +54,7 @@ struct ws_connection_processor {
     struct ws_serializer* serializer; //!< @protected serializer to use
     ev_io dispatcher; //!< @protected dispatching watcher
     ev_prepare flusher; //!< @protected flushing watcher
-    bool is_init; //!< @protected flag indicating whether it's initialized
+    bool is_init; //!< @protected flag indicating whether it's started
```
matthiasbeyer: Sure you do not want another flag for this instead of using the `is_init` flag?

neithernut: As hinted already, _renaming_ it would kind of make sense, but having the running flag lying around does not. The flag as it was used before is absolutely useless.

matthiasbeyer:
ACK. Waiting for CI.

Issue #629, opened 2014-12-20
Kill off zombies


neithernut:
Our exec-utility forks and executes processes, which end up as children of waysome. If those processes terminate, they end up as zombie processes which are attached to waysome. Killing off those zombie-processes is not yet done. Doing it is failry simple. You have to:
- Add an ev-watcher for `SIGCHILD`
- Use `waitpid` or `waitid` to "query" the child process' pid/status (in a non-blocking manner)
- _Optionally_ do something with the info (like print a debug message).

PR #653, opened 2014-12-27
Add child exit handling

matthiasbeyer:
Targets #629.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -56,6 +56,18 @@ handle_sig(
     ev_unloop(loop, EVUNLOOP_ALL);
 }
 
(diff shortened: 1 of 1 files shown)


neithernut:
ACK, but @TheNeikos should have a look at it. Maybe he wants something different.

matthiasbeyer:
We actually discussed it and we came to the conclusion that this is the solution we want. We know that there are benefits from storing the client as abstract "thing" somewhere, but we agreed that this is a) more simple and b) less faulty.

neithernut:
I'll merge this then.

Issue #652, opened 2014-12-27
Cannot connect() from either Python nor Ruby to Unix socket


matthiasbeyer:
I tried the following:

in Ruby:

``` ruby
require 'socket'
s = UNIXSocket.new("/run/user/1000/waysome.sock")
Errno::ECONNREFUSED: Connection refused - connect(2) for "/run/user/1000/waysome.sock"
        from (irb):4:in `initialize'
        from (irb):4:in `new'
        from (irb):4
        from /usr/bin/irb:11:in `<main>'
```

and Python:

``` python
>>> import socket
>>> import sys
>>> import os
>>> s = '/run/user/1000/waysome.sock'
>>> sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
>>> sock.bind(s)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  OSError: [Errno 98] Address already in us
```

So it seems that something is broken. I ran waysome in the debugger for the Ruby
approach, but the `create_connection_cb` isn't called.

neithernut:
You should check whether:
- The accepting callback is set up correctly **using the debugger** (e.g. whether `socket()`, `bind()` and `listen()` are called)
- The handler for accepting is called
- The return value of `accept()` or whether it's even called at all

matthiasbeyer:
I re-checked this stuff.

It works whenever the socket file does not exist before starting waysome. So it seems that the cleanup does not work properly. I will try fix this now.

PR #654, opened 2014-12-27
Fix ws_socket to be able to connect to it

matthiasbeyer:
"Kind of" WIP, as waysome quits (gracefully) after the connection got
closed by the client. I don't know why, yet.

Targets #652.


neithernut:
This is probably because of some signal being emitted. Anyways, the probability of the cause for this bug to be in this piece of code is ... low.

---

ACK.

PR #655, opened 2014-12-28
Add support for custom error strings in ws_errno_tostr()

matthiasbeyer:
Feel free to suggest more changes.

### DIFF
--- a/src/action/manager.c
+++ b/src/action/manager.c
@@ -416,9 +416,10 @@ run_transaction(
     // run processor
     res = ws_processor_exec(&proc);
     if (res < 0) {
-        const char* errmsg = ws_errno_tostr(res);
+        char* errmsg = ws_errno_tostr(res);
         retval = (struct ws_reply*)
                  ws_error_reply_new(transaction, -res, errmsg, NULL);
+        free(errmsg);
         goto cleanup_processor;
     }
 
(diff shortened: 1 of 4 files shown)


neithernut:
This is _nasty_.

matthiasbeyer:
Why?

---

I'd happily change the behaviour if you tell me how to find out what errno number is the highest...

neithernut:
I doubt that custom errnos are of _any_ use here at all. Extending the errnos in an executable is a _bad_ idea. Using the logging facility directly is preferable.
If you want extensible error reporting from functions' return codes, use a language which supports exceptions.

matthiasbeyer:
And if you want to return errors to the plugin?

neithernut:
Returning error codes which are an _expansion_ or errnos is even worse.
You set an invalid error code (e.g. `0`) in the error reply and attach the error string as you usually would.

matthiasbeyer:
So how do we then provide a unique way to find error explanations? Meaning a single instance where the error strings are hardcoded, so we do not have code duplication...

neithernut:
With another system which does **not** interfere with `errno`! E.g. a second function which delivers error strings/formats for enums.
But given an error case which can _not_ be described using a standart-errno-string, the probabilty to have this twice is quite low, so the amount of duplicated code is _minimal_. Hence, I'd close this unmerged.

matthiasbeyer:
ACK for closing this unmerged.

